PR #8, opened 2017-05-12
Smoothie Chart View

TrentHouliston:
Adding the basic smoothie chart view.
- [ ] Get canvas to respond correctly to size changes
- [ ] Create model for displaying data
- [ ] Create sidebar for selecting which data to display

### DIFF
--- a/src/client/components/chart/view.tsx
+++ b/src/client/components/chart/view.tsx
@@ -1,7 +1,73 @@
 import * as React from 'react'
+import {SmoothieChart, TimeSeries} from 'smoothie'
+import * as style from './style.css'
 
-export const Chart = () => (
-    <h1>Chart</h1>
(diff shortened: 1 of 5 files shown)


src/client/components/chart/view.tsx
```diff
@@ -1,7 +1,100 @@
...
+    this.timeseries = []
+  }
+
+  public render(): JSX.Element {
+    const sidebarContent = <b>Hello!</b>
+    return (
+      <div className={style.chart}>
+        <MenuBar/>
+        <div className={style.chart__canvasContainer}>
+          <Sidebar sidebar={sidebarContent}
```
BrendanAnnable: This sidebar needs to exist outside the `chart__canvasContainer`

PR #16, opened 2017-05-13
[WIP] Scatterplot View

Phasedd:
Adding a scatterplot view.


declarations/plotly.js/index.d.ts
```diff
@@ -3,8 +3,11 @@ import 'plotly.js'
 declare module 'plotly.js' {
   type Update = { x: Array<number[]>, y: Array<number[]> }
   type ShortLayout = { width: number, height: number }
+  type ShortMarker = {size: number}
+  type ShortData = {x: Array<number[]>, y: Array<number[]>, z: Array<number[]>, mode: string, type: string, marker: ShortMarker, name: string}
```
BrendanAnnable: I believe `Array<number[]>` is using two different syntaxes, stick with one and just use `number[][]` 🙂

Phasedd: was mostly sticking to what @MonicaOlejniczak used.

BrendanAnnable: yeah I think they should both be changed 🙂

MonicaOlejniczak: Oh does that work now? I used to get errors using the short-hand way for multi-dimensional arrays at my previous job. Wonder if that was an issue with the old school typings or the compiler.

BrendanAnnable:
I will get to this shortly... could you merge master into this branch in the meantime? 🙂

Phasedd:
Master was merged a few days ago before I started working on it again, nothing has been committed to master since.

It is not completely ready for merging into master yet, the only thing that stops it is to remove unnecessary traces from being drawn. The code supports preventing traces from being draw until it is toggled on, but at the moment it will add the trace to the graph as soon as a new data set is seen (just need to change some bools for it to not be added as soon as a new data set is seen).

plotly.js does support a way to toggle traces on and off, but it clutters the UI a lot. To get around that in NUsight1 there is a selection UI like for the smoothie charts, but with more options then just toggling it on and off.

src/client/components/scatter_plot/view.tsx
```diff
@@ -1,5 +1,256 @@
...
+  addTrace: boolean
+}
+
+interface DataPoint {
+  label: string
+  values: number[]
+}
+
+@observer
+export class ScatterplotView extends React.Component<any, any> {
```
BrendanAnnable: This component needs some refactoring to be inline with some guidelines which I haven't done a good job of documenting yet 🙂 

Some goals I had when writing the localisation component:

  - If possible, they should just render, only using the props they were given.
  - Preferably they have very little "logic".
  - Generally this means they become almost "stateless", i.e. there are a minimal number of member fields in the view itself.
  - Controllers were just responsible for responding to user actions, and appropriately manipulating the state. They know nothing about the DOM, e.g. they should not be given an HTML element for instance, or directly call `window.someGlobalThing`.
  - If you take [`LocalisationView`](https://github.com/NUbots/NUsight2/blob/master/src/client/components/localisation/view.tsx) as an example, it is only concerned about how things look, and propagating view events (`onClick`, etc) out to the controller. Views never directly manipulates data.

These ideas come down to some rules of thumb which roughly align with standard [MVC](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller):
- Views should as stateless.
- Models should contain all your state.
- Controllers should do all your logic.

To achieve this for the scatterplot component, I think we would need to:
- Moving all your state, e.g. `traces`, `nextTraceId`, `graphUpdateX`, `graphUpdateY`, `graphUpdateZ`, `maxPoints`, `fps`, into a separate `ScatterPlotModel`.
- Moving any logic into the controller
- Remove any view/dom related things from the controller, instead manipulate the state which should trigger a view change.

src/client/components/scatter_plot/view.tsx
```diff
@@ -1,5 +1,256 @@
...
+
+      me.graphUpdateX.clear()
+      me.graphUpdateY.clear()
+      me.graphUpdateZ.clear()
+    }, 1000 / this.fps)
+
+    this.simulateData()
+  }
+
+  public simulateData(): void {
```
BrendanAnnable: What is the long-term plan for this component? What sorts of data do we plan on displaying?

Phasedd: The scatterplots are just a different way to view the same data that smoothie charts display.

BrendanAnnable: Should they maybe share the same model then?

Phasedd: I think the plotly model may be a lot more complex then the smoothie model.

BrendanAnnable: In what way, more data?

Phasedd: Each trace is going to require more properties compared to smoothie I believe.

BrendanAnnable: What kind of properties? (I don't know much about either)

Phasedd: I am not too sure about smoothie, I think it is really only label and data.

Plotly how ever requires x, y, z, marker info (size, colour, icon etc), type (scatterplot, line graph, contour, histogram etc), hover info, label, id (the id that plotly assigns the trace so we can add data to it), and most likely even more.

The plotly graph it self has some specific stuff as well, such as controlling the axis min and max if we want to be able to restrict the graph to a certain area.

src/simulators/chart_simulator.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export class ChartSimulator implements Simulator {
+  public static of() {
+    return new ChartSimulator()
+  }
+
+  public simulate(time: number): Message[] {
+    const messageType = 'message.support.nubugger.DataPoint'
+
+    const now = time
```
Phasedd: oh right, I did that to save time on testing to make sure it worked, but forgot to update it.

src/server/dev.ts
```diff
@@ -12,7 +12,9 @@ import webpackConfig from '../../webpack.config'
 import { RobotSimulator } from '../simulators/robot_simulator'
 import { SimulatorStatus } from '../simulators/robot_simulator'
 import { SensorDataSimulator } from '../simulators/sensor_data_simulator'
+import { ChartSimulator } from '../simulators/chart_simulator'
 import { NUSightServer } from './app/server'
+import { Chart } from '../client/components/chart/view'
```
Phasedd: hmm, yep, must of been IntelliJ auto importing when I wrote Chart instead of ChartSimulator or something.

src/shared/proto/Matrix.proto
```diff
@@ -1,105 +1,105 @@
...
-    ivec4 t = 4;
-}
-
-message umat44 {
-    uvec4 x = 1;
-    uvec4 y = 2;
-    uvec4 z = 3;
-    uvec4 t = 4;
-}
+syntax = "proto3";
```
Phasedd: I haven't touched the proto files, I think it did convert to CRLF, but I told IntelliJ to do LF instead before committing for files and line separators.

src/client/components/scatter_plot/model.ts
```diff
@@ -0,0 +1,139 @@
...
+      maxPoints: 100,
+      traces: new Map<string, Trace>(),
+      graphUpdateX: new Map<number, number[]>(),
+      graphUpdateY: new Map<number, number[]>(),
+      graphUpdateZ: new Map<number, number[]>(),
+      graphType: 'scatterplotgl',
+    })
+  }
+
+  @action
```
BrendanAnnable: I haven't had a chance yet to think about this enough to give a proper suggestion, but do you think its possible to keep all `@action` annotations within the controller? This lets us keep consistent with having all model mutations within controllers.

(p.s. `network.ts` is an pseudo-exception to this rule, since it is essentially a network-only controller just for reacting to network events, rather than browser events).

PR #17, opened 2017-05-13
Fix localisation canvas autosizing

BrendanAnnable:
This fixes the localisation window such that is shrinks to size when the window is made smaller 😄

### DIFF
--- a/src/client/components/localisation/style.css
+++ b/src/client/components/localisation/style.css
@@ -42,8 +42,15 @@
 	border: 1px solid rgba(0, 0, 0, 0.4);
 }
 
+.localisation__canvasContainer {
+  flex-grow: 1;
+  position: relative;
+}
+
 .localisation__canvas {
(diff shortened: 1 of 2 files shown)


src/client/components/localisation/style.css
```diff
@@ -42,8 +42,15 @@
...
+.localisation__canvasContainer {
+  flex-grow: 1;
+  position: relative;
+}
+
 .localisation__canvas {
-	flex-grow: 1;
+  width: 100%;
+  height: 100%;
+  position: absolute;
```
BrendanAnnable: When a child is position absolute and the parent is position relative, coordinates such as width and height become relative to the parent instead of the page.

Essentially the canvas is forced to take the shape of its parent, which flows nicely within the flexbox layout.

This gives us exactly what we want, an autosizing canvas element which flows with the natural page layout!

PR #18, opened 2017-05-15
Allow the use of global css files without mangling them

TrentHouliston:


webpack.config.ts
```diff
@@ -71,6 +74,14 @@ export default {
           ],
         }),
       },
+      // global css
```
BrendanAnnable: Maybe extend this comment to be more useful? 🙂 e.g.

```
/* External libraries generally do not support css modules, so the selector mangling will break external components. This separate simplified loader is used for anything within the node_modules folder instead. */
```

PR #19, opened 2017-05-18
[WIP] Configuration View

JosephusPaye:
Adding the configuration view. Feedback welcome.


src/client/components/configuration/style.css
```diff
@@ -0,0 +1,48 @@
+.configuration {
+	flex-grow: 1;
+	flex-shrink: 1;
+	display: flex;
+	flex-direction: column;
+    font-weight: normal;
```
BrendanAnnable: Could you reformat all these CSS files to use 2 space indentation? 🙂 I think its my fault for checking in some non-2-space files.

src/client/components/configuration/tree/tree.tsx
```diff
@@ -0,0 +1,68 @@
...
+import FolderIcon from './folder.svg'
+import * as style from './style.css'
+
+export interface Node {
+  label: string,
+  expanded: boolean,
+  leaf: boolean,
+  selected: boolean,
+  children?: Node[]
+}
```
BrendanAnnable: You can remove all the commas from these interface definitions, e.g.

```
export interface Node {
  label: string
  expanded: boolean
  leaf: boolean
  selected: boolean
  children?: Node[]
}
```

I'd put it in a tslint rule if there was one (https://github.com/palantir/tslint/issues/960)...

src/client/components/configuration/view.tsx
```diff
@@ -0,0 +1,71 @@
...
+      <div className={style.configuration__content}>
+        <div className={style.configuration__contentPlaceholder}>
+          Select a file to edit
+        </div>
+      </div>
+    </div>
+  </div>
+)
+
+export default Configuration
```
BrendanAnnable: I'm going to be removing all default exports, so might as well remove this one 😄

src/client/components/configuration/tree_store.tsx
```diff
@@ -0,0 +1,39 @@
...
+  @observable
+  public selectedNode: Node
+
+  constructor(data: Node) {
+    this.data = data
+    this.onNodeClick = this.onNodeClick.bind(this)
+  }
+
+  @action
+  public onNodeClick(node: Node): void {
```
BrendanAnnable: One of my architectural goals is to keep models as dumb as possible, such that they are _just_ data and no behaviour.

That means actions like these need to live somewhere else, preferably somewhere that we can test.

This is why I created a `LocalisationPresenter` for the localisation component, and any `@action` lives as a method on the presenter, such that the `LocalisationModel` itself does not have any actions on it. I'll soon be renaming `LocalisationPresenter` to `LocalisationController` as well.

Could we follow a similar pattern here? 😄  You'll probably want a separate `ConfigurationController` class where these actions can live 🙂

src/client/components/app/style.css
```diff
@@ -1,5 +1,9 @@
 @import url('https://fonts.googleapis.com/css?family=Open+Sans:300,400,500');
 
+* {
+	box-sizing: border-box;
```
BrendanAnnable: Do you think this will ever come back to bite us? Would libraries ever assume `content-box` is the default and break? Maybe its fine and we'll revisit if its a problem.

src/client/components/configuration/tree/file.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="16" height="16" fill="#999" viewBox="0 0 24 24">
+    <path d="M13,9H18.5L13,3.5V9M6,2H14L20,8V20A2,2 0 0,1 18,22H6C4.89,22 4,21.1 4,20V4C4,2.89 4.89,2 6,2M11,4H6V20H11L18,20V11H11V4Z" />
+</svg>
```
BrendanAnnable: Can you ensure all these SVG files have a trailing newline? 🙂

src/client/components/configuration/tree/tree.tsx
```diff
@@ -0,0 +1,68 @@
...
+      </div>
+    )
+  }
+
+  private onClick = (e: any) => {
+    this.props.onClick(this.props.data)
+  }
+}
+
+export default Tree
```
BrendanAnnable: I'm going to be removing all default exports, so might as well remove this one 😄

src/client/components/configuration/style.css
BrendanAnnable: Probably should look into adding something like https://github.com/stylelint/stylelint

JosephusPaye: Will do.

src/client/components/configuration/tree/tree.tsx
JosephusPaye: Gotcha.

src/client/components/app/style.css
JosephusPaye: Another options is this, which allows for easily resetting the `box-sizing`:

```css
html {
  box-sizing: border-box;
}

*, *:before, *:after {
  box-sizing: inherit;
}
```

Reference: https://css-tricks.com/box-sizing/#article-header-id-6

JosephusPaye:
@BrendanAnnable Thanks for the suggestions, I'll make the requested changes and push updates soon.

src/client/components/configuration/tree_store.tsx
JosephusPaye: Sounds good, will do.

src/client/components/configuration/tree/tree.tsx
```diff
@@ -0,0 +1,66 @@
...
+                (this.props.data.expanded ? <FolderIconOpen /> : <FolderIcon />) :
+                <FileIcon />
+            }
+          </div>
+
+          <div className={style.treenode__label}>{ this.props.data.label }</div>
+        </div>
+
+        <div className={style.treenode__children}>
+          { children.map((child, i) => <Tree key={i} data={child} level={level + 1} onClick={this.props.onClick} />) }
```
BrendanAnnable: Are we safe to use `i` as a key here?

Source: https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318

src/client/components/configuration/tree/tree.tsx
```diff
@@ -0,0 +1,66 @@
...
+              hasChildren ?
+                (this.props.data.expanded ? <FolderIconOpen /> : <FolderIcon />) :
+                <FileIcon />
+            }
+          </div>
+
+          <div className={style.treenode__label}>{ this.props.data.label }</div>
+        </div>
+
+        <div className={style.treenode__children}>
```
BrendanAnnable: Can we optimise this by only rendering children if it is expanded? Then we can delete the higher specificity CSS selectors like `.treenode--expanded > .treenode__children`

src/client/components/configuration/tree/tree.tsx
```diff
@@ -0,0 +1,66 @@
...
+          { children.map((child, i) => <Tree key={i} data={child} level={level + 1} onClick={this.props.onClick} />) }
+        </div>
+      </div>
+    )
+  }
+
+  private onClick = (e: any) => {
```
BrendanAnnable: You can use `MouseEvent` instead of `any` here

src/client/components/configuration/view.tsx
```diff
@@ -0,0 +1,69 @@
+import * as React from 'react'
+import * as style from './style.css'
+import { Node, Tree } from './tree/tree'
+import { TreeStore } from './tree_store'
+
+const store = new TreeStore({
```
BrendanAnnable: I think its best that we stop using the word `store` and `model` and just use one through throughout the code base. Then we can maybe think of it as a more traditional "model view controller" architecture. So I suggest we replace every occurrence of `store` with `model`.

src/client/components/configuration/view.tsx
```diff
@@ -0,0 +1,69 @@
...
+            selected: false,
+            children: [] as Node[],
+        },
+    ],
+})
+
+export const Configuration = () => (
+  <div className={style.configuration}>
+    <div className={style.configuration__header}>
+      <h1 className={style.configuration__headerTitle}>Configuration</h1>
```
BrendanAnnable: We'll have to make a component for this header stuff 🙂

src/client/components/configuration/tree/tree.tsx
```diff
@@ -0,0 +1,66 @@
...
+}
+
+export interface TreeProps {
+  data: Node
+  level?: number
+  onClick(node: Node): void
+}
+
+@observer
+export class Tree extends React.Component<TreeProps, any> {
```
BrendanAnnable: The second param type `any` should be `void` I believe

src/client/components/configuration/tree/tree.tsx
```diff
@@ -0,0 +1,66 @@
...
+  constructor(props: any, context: any) {
+    super(props, context)
+    this.onClick = this.onClick.bind(this)
```
BrendanAnnable: This is not needed when you use class properties (which you're already doing!), so you can delete this line 🙂 

Normal class method, needs the `.bind` in constructor if you want to pass it directly to something which won't keep the context:

```
onClick(e: MouseEvent) {
  // "this" might change
}
```

Arrow function class property, automatically binds for you:
```
onClick = (e: MouseEvent) => {
  // "this" will always be the class instance
}
```

src/client/components/configuration/tree/tree.tsx
BrendanAnnable: Maybe something like:

```
        {this.props.data.expanded &&
          <div className={style.treenode__children}>
            ...
          </div>
        }
```

src/client/components/configuration/tree/tree.tsx
```diff
@@ -0,0 +1,66 @@
...
+    const classes = classnames(
+      style.treenode,
+      { [style['treenode--expanded']]: this.props.data.expanded },
+      { [style['treenode--selected']]: this.props.data.selected },
+    )
+    const headerInlineStyle = { paddingLeft: 8 + (level * 22) + 'px' }
+
+    return (
+      <div className={classes}>
+        <div className={style.treenode__header} onClick={this.onClick} style={headerInlineStyle}>
```
BrendanAnnable: Could we use `ul` and `li` for this HTML structure so we don't need to use a calculated `paddingLeft`?

http://jsbin.com/zoyabovami/edit?html,css,output

Credit to @MonicaOlejniczak for the idea 🙂

src/client/components/configuration/tree/tree.tsx
JosephusPaye: Using `MouseEvent` gives this error:

```shell
ERROR in [at-loader] ./src/client/components/configuration/tree/tree.tsx:44:14
    TS2322: Type '{ className: any; onClick: (e: MouseEvent) => void; style: { paddingLeft: string; }; }' is not assignable to type 'HTMLProps<HTMLDivElement>'.
  Types of property 'onClick' are incompatible.
    Type '(e: MouseEvent) => void' is not assignable to type 'EventHandler<MouseEvent<HTMLDivElement>> | undefined'.
      Type '(e: MouseEvent) => void' is not assignable to type 'EventHandler<MouseEvent<HTMLDivElement>>'.
        Types of parameters 'e' and 'event' are incompatible.
          Type 'MouseEvent<HTMLDivElement>' is not assignable to type 'MouseEvent'.
            Property 'fromElement' is missing in type 'MouseEvent<HTMLDivElement>'.
```

src/client/components/configuration/tree/tree.tsx
JosephusPaye: Fixed.

src/client/components/configuration/tree/tree.tsx
JosephusPaye: Nice, didn't know that!

src/client/components/configuration/tree/tree.tsx
JosephusPaye: Done!

src/client/components/configuration/tree/tree.tsx
JosephusPaye: Wasn't sure what to use as the key, and React was complaining without it, that's why I used the index.

We could use `label` + `level`, but that could be duplicated.

src/client/components/configuration/view.tsx
JosephusPaye: Done!

src/client/components/configuration/tree/tree.tsx
JosephusPaye: I had that initially (see below) but changed to `paddingLeft` because I wanted the hover and selected highlights to be full width. I can switch back if necessary.

Using `ul`/`li`:



Using `div` and `paddingLeft`:

src/client/components/configuration/view.tsx
JosephusPaye: I've made two wrapper components [view](https://github.com/NUbots/NUsight2/tree/feature/configuration/src/client/components/configuration/view) (for top-level views like dashboard or localization) and [panel](https://github.com/NUbots/NUsight2/tree/feature/configuration/src/client/components/configuration/panel) (for stuff like sidebars) which could be useful.

I've placed them under [`components/configuration`](https://github.com/NUbots/NUsight2/tree/feature/configuration/src/client/components/configuration) for now, but I think they should go higher (`components/view|panel/` or `components/common/view|panel/`?) since they could be used by other views.

I can extract them into a separate pull request if needed.

src/client/components/configuration/tree/tree.tsx
MonicaOlejniczak: Hmmm, good point! You can sorta do it with pure CSS but it's pretty hacky.

BrendanAnnable: Ah! Makes sense. Maybe add a comment explaining why then? 🙂 Otherwise the next guy will think the same thing haha

BrendanAnnable: We can still use `ul` and `li` though at least for the semantics

JosephusPaye: Done: https://github.com/NUbots/NUsight2/pull/19/commits/6e431aca1c27000b468e78422d82fdb037bd99cf

JosephusPaye:
Hey @BrendanAnnable,

After pulling in the latest master (https://github.com/NUbots/NUsight2/commit/19929686d18e6d143e63cb3ec54f9d56b1926611), I'm having trouble compiling the code. 

<details>
<summary>Errors</summary>

```shell
ERROR in [at-loader] ./src/client/components/configuration/editor/editor.tsx:33:2
    TS2345: Argument of type 'typeof Editor' is not assignable to parameter of type 'string[]'.
  Property 'includes' is missing in type 'typeof Editor'.

ERROR in [at-loader] ./src/client/components/configuration/editor/editor.tsx:39:41
    TS2605: JSX element type 'MapField' is not a constructor function for JSX elements.

ERROR in [at-loader] ./src/client/components/configuration/editor/editor.tsx:40:42
    TS2605: JSX element type 'ListField' is not a constructor function for JSX elements.

ERROR in [at-loader] ./src/client/components/configuration/editor/editor.tsx:45:11
    TS2605: JSX element type 'ScalarField' is not a constructor function for JSX elements.

ERROR in [at-loader] ./src/client/components/configuration/editor/fields/scalar_field.tsx:22:13
    TS2605: JSX element type 'NumberValue' is not a constructor function for JSX elements.
  Types of property 'setState' are incompatible.
    Type '{ <K extends never>(f: (prevState: void, props: FieldProps) => Pick<void, K>, callback?: (() => a...' is not assignable to type '{ <K extends never>(f: (prevState: {}, props: any) => Pick<{}, K>, callback?: (() => any) | undef...'.
      Types of parameters 'f' and 'f' are incompatible.
        Type '(prevState: {}, props: any) => Pick<{}, any>' is not assignable to type '(prevState: void, props: FieldProps) => Pick<void, any>'.
          Types of parameters 'prevState' and 'prevState' are incompatible.
            Type 'void' is not assignable to type '{}'.
```
</details>

<br>
It seems to have something to do with `observer` from `mobx-react`.

I get no errors when I run `yarn tslint`. To test see my latest commit: https://github.com/NUbots/NUsight2/pull/19/commits/9135187c61b1b0f6e9a072e58070db1576b3f73a.

PR #20, opened 2017-05-18
Add .editorconfig

JosephusPaye:
EditorConfig (http://editorconfig.org/) allows for configuring code style conventions across different editors. The file I've added ensures that indents are 2 spaces and files use Unix-style (LF) line endings (these can be changed).

I use Sublime on Windows at the moment (downloading IntelliJ @TrentHouliston 😄)  and this allows me to automatically have Unix-style line endings and consistent indents.

EditorConfig is already supported by IntelliJ IDEA and there are plugins (http://editorconfig.org/#download) for most editors/IDEs.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,7 +3,6 @@
 # This is the top-most EditorConfig file
 root = true
 
-# Unix-style newlines with indentation of 2 spaces
 [*]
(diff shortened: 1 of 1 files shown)


JosephusPaye:
Added!

PR #21, opened 2017-05-19
Make nodemon ignore client-side changes

BrendanAnnable:
Currently nodemon will restart the entire server when you make client-side changes, i.e. edit a file in `src/client`.

This is unneeded and makes development slow! So this PR makes nodemon ignore both tests and anything in `src/client`.

### DIFF
--- a/nodemon.json
+++ b/nodemon.json
@@ -1,0 +1,7 @@
+{
+  "verbose": true,
+  "ignore": ["*.tests.ts", "src/client"],
+  "execMap": {
+    "ts": "ts-node"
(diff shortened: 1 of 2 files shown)


nodemon.json
```diff
@@ -0,0 +1,7 @@
+{
+  "verbose": true,
+  "ignore": ["*.tests.ts", "src/client"],
+  "execMap": {
+    "ts": "ts-node"
```
BrendanAnnable: Associates `*.ts` files with `ts-node`

PR #24, opened 2017-05-19
Clean up tsconfig and target ES6

BrendanAnnable:
Clean up the tsconfig, e.g removed `build` since it doesn't exist.

Also targeting es6 instead of es5 because we don't need to support old browsers. Should improve runtime performance and reduce the size of the overall JS.


BrendanAnnable:
I broke the build 😛 please review again.

Now that we're targeting ES6, classes need to be declared in order or use. Putting my hacky models in separate files solves the issue.

BrendanAnnable:
Abandoning targeting ES6 for now due to build failures

https://github.com/NUbots/NUsight2/pull/25

PR #31, opened 2017-05-19
Add husky prepush hook

BrendanAnnable:
Automatically runs `yarn tslint` on `git push`. It is fast enough not to be annoying and you can use `--no-verify` to bypass if need.

This would have saved me a few times already since it doesn't error during development with just `yarn start` and only found out when the build failed 😛 

p.s. you can run `yarn tslint:fix` to automatically fix most tslint issues!


BrendanAnnable:
@Phasedd @JosephusPaye please also approve if this works for you 🙂

BrendanAnnable:
@JosephusPaye ping 🙂

JosephusPaye:
Hey @brendanAnnable, sorry for not responding earlier, been swamped with uni work. 

It think it works for me:



However, should `tslint` also be setup to lint `.tsx` files also? I have to manually run this for actual linting to work for me:

```shell
tslint -c tslint.json 'src/**/*.ts?(x)'
```

BrendanAnnable:
No worries @JosephusPaye — I completely understand!

Great catch regarding `*.tsx`, I'll fix that up.

BrendanAnnable:
Done!

JosephusPaye:
Also it seems like the single quotes doesn't work on Windows (when ran through `yarn` or `npm run`).

BrendanAnnable:
Hmm really? Do you need double quotes?

JosephusPaye:
It works without quotes.

BrendanAnnable:
Without quotes means we're passing off the path globbing to the specific shell of the user. I would have expected no quotes to actually be less consistent across platforms than with quotes... maybe you need to run `yarn` again and make sure you have installed the new dependencies?

JosephusPaye:
I ran `git pull` and then this, with no changes:

BrendanAnnable:
Hmm yep seeing the same issue on my windows VM, double quotes also work but I'll look into it a bit further before settling on anything.

BrendanAnnable:
Switched to use `{,x}` instead of `?(x)` as it was the parenthesis causing the windows issue. Tested on OSX and Windows.

BrendanAnnable:
All good to merge @JosephusPaye? 🙂

JosephusPaye:
All good now.

BrendanAnnable:
Hurray!

PR #36, opened 2017-05-26
Make localisation controller stateless

BrendanAnnable:
Making controllers stateless should solve our composition problems. State should instead live in the model.

This allows a single controller to be shared among all instances of a view and allow us to [dependency inject](https://github.com/NUbots/NUsight2/pull/38) the controller into a view. This lets us use normal JSX throughout the app, without worrying about instantiating and destroying the corresponding controllers. It also makes testing the controllers trivial.

In the case of localisation, we won't see any improvement as there is only a single instance of the view anyway, but in the case of other components which display a list of `<MyItem/>`, they can each use a corresponding `MyItemController` much more easily. Hopefully this makes more sense with more examples.

The controllers then become somewhat analogous to reducers in a redux architecture. The difference being that we don't return a new state, we just mutate the state. Also, we don't have to be synchronous. Actions can kick off asynchronous tasks which may eventually mutate the state.


src/client/components/localisation/view.tsx
```diff
@@ -1,16 +1,24 @@
...
-@inject('localisationStore')
+interface LocalisationViewProps {
+  model: LocalisationModel
+}
+
 @observer
-export class LocalisationView extends React.Component<any, any> {
+export class LocalisationView extends React.Component<LocalisationViewProps, any> {
+  @inject(LocalisationController)
+  private presenter: LocalisationController
```
MonicaOlejniczak: Rename to controller

src/client/components/localisation/view.tsx
```diff
@@ -67,59 +74,58 @@ export class LocalisationView extends React.Component<any, any> {
 
   private onAnimationFrame = (time: number) => {
     this.rafId = requestAnimationFrame(this.onAnimationFrame)
-    this.props.presenter.onAnimationFrame(time)
+    this.presenter.onAnimationFrame(this.props.model, time)
```
MonicaOlejniczak: Rename

src/client/components/localisation/view.tsx
```diff
@@ -67,59 +74,58 @@ export class LocalisationView extends React.Component<any, any> {
...
     this.renderer.render(viewModel.scene, viewModel.camera)
 
-    runInAction(() => model.time.lastRenderTime = model.time.time)
+    runInAction(() => this.props.model.time.lastRenderTime = this.props.model.time.time)
   }
 
   private onClick = (e: MouseEvent) => {
     if (e.button === 0) {
-      this.props.presenter.onLeftClick(this)
+      this.presenter.onLeftClick(this.props.model, this)
```
MonicaOlejniczak: Rename

PR #38, opened 2017-05-26
[WIP] Add InversifyJS

BrendanAnnable:


src/client/components/localisation/view.tsx
```diff
@@ -3,19 +3,22 @@ import { runInAction } from 'mobx'
 import { observer } from 'mobx-react'
 import * as React from 'react'
 import { WebGLRenderer } from 'three'
+import { lazyInject } from '../../../inversify.config'
```
MonicaOlejniczak: Add a webpack alias for this. Can it also have a better name while you're at it? I'm not fond of it being called a config.

src/client/components/localisation/view.tsx
```diff
@@ -3,19 +3,22 @@ import { runInAction } from 'mobx'
...
 
 interface LocalisationViewProps {
-  model: LocalisationModel,
-  presenter: LocalisationPresenter
+  model: LocalisationModel
 }
 
 @observer
 export class LocalisationView extends React.Component<LocalisationViewProps, any> {
+  @lazyInject(LocalisationPresenter)
```
MonicaOlejniczak: Are we always going to lazy inject? I feel like we shouldn't know the way it's being injected in the localisation view.

src/client/components/localisation/view.tsx
BrendanAnnable: Coming back to this in another PR! Name suggestions welcome, but `inversify.config` is the standard/recommended.

src/client/components/localisation/view.tsx
BrendanAnnable: Fixed

PR #39, opened 2017-05-27
Inject LocalisationModel into LocalisationView

BrendanAnnable:
Removed the `Provider` component provided by `mobx-react` in favour of using the dependency injection provided by `inversify`.

### DIFF
--- a/src/client/components/localisation/view.tsx
+++ b/src/client/components/localisation/view.tsx
@@ -10,14 +10,13 @@ import { ViewMode } from './model'
 import * as style from './style.css'
 import { LocalisationViewModel } from './view_model'
 
-interface LocalisationViewProps {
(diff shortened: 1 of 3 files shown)


src/client/components/localisation/view.tsx
```diff
@@ -10,15 +10,14 @@ import { ViewMode } from './model'
 import * as style from './style.css'
 import { LocalisationViewModel } from './view_model'
 
-interface LocalisationViewProps {
-  model: LocalisationModel
-}
-
 @observer
-export class LocalisationView extends React.Component<LocalisationViewProps, any> {
+export class LocalisationView extends React.Component<{}, any> {
```
MonicaOlejniczak: Change to HTMLProps<JSX.Element>

PR #42, opened 2017-05-28
Sensor data simulator

BrendanAnnable:
This PR:
- Adds a general network simulation system, such that it should be relatively easy to add new simulators.
- Add a specific simulator for sensor data. Currently just simulates servo positions to follow a walk-like movement.

Running the server + simulators:
- `yarn run start:sim`
This runs the simulators using a fake in-memory NUClearNet. This should be suitable for normal development and can be run completely offline with no overhead.

Running just the simulators (not usually necessary):
- `yarn run simulate` / `yarn run simulate -- --all`
- `yarn run simulate -- --sensors`
These run the simulators using the real NUClearNet, as they are run in a separate binary. These should be essentially indistinguishable from real robots to NUsight.

This PR does not:
- Add any networking between the simulators and NUsight (probably the next PR will do that).


.travis.yml
```diff
@@ -6,6 +6,14 @@ node_js:
...
   - '7'
   - '8'
+env:
+  - CXX=g++-4.9
+addons:
+  apt:
+    sources:
+      - ubuntu-toolchain-r-test
+    packages:
+      - g++-4.9
```
BrendanAnnable: For compiling `nuclearnet.js`

package.json
```diff
@@ -3,17 +3,19 @@
...
-  "main": "index.js",
   "scripts": {
     "test": "jest",
     "test:ci": "yarn test -- --ci",
     "test:watch": "yarn test -- --watch",
     "coverage": "yarn test -- --coverage",
     "coverage:ci": "yarn test:ci -- --coverage",
-    "tslint": "tslint -c tslint.json 'src/**/*.ts{,x}'",
-    "tslint:fix": "tslint -c tslint.json 'src/**/*.ts{,x}' --fix",
+    "tslint": "tslint -c tslint.json 'src/**/*.ts{,x}' --exclude 'src/shared/proto/**'",
```
BrendanAnnable: Without this tslint takes _much_ longer trying to lint the generated `messages.js` (which doesn't conform to our style guide).

src/server/dev.ts
```diff
@@ -37,3 +46,16 @@ server.listen(port, () => {
...
+if (withSimulators) {
+  const robotSimulator = new RobotSimulator({
+    name: 'Sensors Simulator',
+    network: FakeNUClearNet.of(),
+    clock: NodeSystemClock,
+    simulators: [
+      SensorDataSimulator.of(),
+    ],
+  })
+  robotSimulator.simulateWithFrequency(60)
```
BrendanAnnable: NUsight itself is not listening to these messages as of yet, but next PR...

tsconfig.json
```diff
@@ -12,6 +12,7 @@
     "noImplicitReturns": false,
     "removeComments": true,
     "strictNullChecks": true,
+    "downlevelIteration": true,
```
BrendanAnnable: Apparently this setting is needed in order to use ES6 `for..of` loops.

tslint.json
```diff
@@ -1,35 +1,42 @@
 {
```
BrendanAnnable: The only change to the file is the addition of `"member-access": true,`, and then I ran reformat...

PR #48, opened 2017-06-17
Add networking

BrendanAnnable:
Haven't thoroughly gone through this, but its enough to unblock everyone and we can iterate.

networking_simulation


src/client/components/localisation/network.ts
```diff
@@ -0,0 +1,45 @@
...
+    this.network.on(Sensors, this.onSensors)
+  }
+
+  public destroy() {
+    this.network.off()
+  }
+
+  @action
+  private onSensors = (sensors: Sensors) => {
+    // TODO (Annable): Remove hardcoded values.
```
TrentHouliston: Isn't this todo done?

src/client/index.tsx
```diff
@@ -17,6 +16,8 @@ import { NUClear } from './components/nuclear/view'
 import { Scatter } from './components/scatter_plot/view'
 import { Subsumption } from './components/subsumption/view'
 import { Vision } from './components/vision/view'
+import { container } from './inversify.config'
+import Sensors = message.input.Sensors
```
TrentHouliston: Unused?

src/client/components/localisation/network.ts
BrendanAnnable: Ah, not quite. It's currently hardcoded to use `this.model.robots[0]`, instead of actually knowing about the connected robots (next PR).

src/client/index.tsx
BrendanAnnable: Removed!

BrendanAnnable:
@TrentHouliston ok this is actually working now 🙂 Had to do quite a few changes, but now things actually work.

i.e. `yarn start` now uses real networking (tested with both `keyboardwalkfake` and `yarn run simulate`), and `yarn start:sim` uses fake networking + runs the simulators.

BrendanAnnable:
Manually deployed to heroku to test prod simulation, its working! 😄 

https://nusight.herokuapp.com/localisation

src/client/components/localisation/network.ts
```diff
@@ -0,0 +1,45 @@
+import { action } from 'mobx'
+import { message } from '../../../shared/proto/messages'
```
TrentHouliston: This should definitely have an alias, we are going to be doing this a lot

src/client/network/global_network.ts
```diff
@@ -0,0 +1,77 @@
...
+    if (listeners) {
+      for (const listener of listeners.values()) {
+        listener(message)
+      }
+    }
+  }
+}
+
+function getMessageTypeName(messageType: MessageType<Message>) {
+  // TODO (Annable): Generate this map.
```
TrentHouliston: This should be done before the PR is merged, otherwise networking is useless for anything that's not sensors!

src/client/network/raw_socket.ts
```diff
@@ -0,0 +1,41 @@
...
+@injectable()
+export class RawSocket {
+  private socket: Socket
+
+  public static of() {
+    return new RawSocket()
+  }
+
+  public on(event: string, fn: (...args: any[]) => void) {
+    this.connectIfNotConnected()
```
TrentHouliston: Isn't this the kind of thing that should be done in the constructor or similar?

src/client/components/localisation/network.ts
BrendanAnnable: The problem an alias is that IntelliJ cannot automatically add the import for you, I'd rather never type an import if I don't have to 😛

src/client/network/global_network.ts
BrendanAnnable: Done.

src/client/network/raw_socket.ts
BrendanAnnable: I avoid constructors doing any kind of work. It makes classes much easier to test because the constructor doesn't attempt to perform any side effects, such as networking.

I could add an explicit `connect()` method, but since I'm making everything else in networking lazy, I think it also makes sense to only connect if someone actually tries to listen for something on the network as well.

BrendanAnnable:
@TrentHouliston changes in

PR #49, opened 2017-06-17
Add a skybox to localisation

Phasedd:
Adds a skybox to the localisation view.

image


src/client/components/localisation/skybox/sky_shader.ts
```diff
@@ -0,0 +1,207 @@
...
+    this.uniforms = {
+      luminance: { value: 1 },
+      turbidity: { value: 2 },
+      rayleigh: { value: 1 },
+      mieCoefficient: { value: 0.005 },
+      mieDirectionalG: { value: 0.8 },
+      sunPosition: { value: new Vector3() },
+    }
+
+    this.vertexShader = [
```
TrentHouliston: These shaders should be in a file and loaded separately rather than placed directly in here!

Phasedd: Ok, I will change it, I thought it would be ok seeing as the files entire job was to just load it, and pretty much no maintenance would be needed.

src/client/components/localisation/skybox/model.ts
```diff
@@ -0,0 +1,48 @@
+import { observable } from 'mobx'
+import { BackSide } from 'three'
+import { UniformsUtils } from 'three'
+import { ShaderMaterial } from 'three'
+import { SphereBufferGeometry } from 'three'
+import { Mesh } from 'three'
```
BrendanAnnable: An import from `three` should be a view model concern.

The model should should be unaware of how it will be used or displayed, and as such, only contain _raw_ data if possible, and the view model should be concerned with creating library-specific data structures and displaying the data (three/plotly/smoothie, etc). Its not always possible for performance reasons, e.g. when mass real-time data is involved. But in this case, most of these should be able to be moved into the view model without too much trouble 🙂 

So I think the only data that should live in this model is the uniform value settings, since they are just numbers.

src/client/components/localisation/skybox/sky_shader.ts
```diff
@@ -0,0 +1,35 @@
+import { observable } from 'mobx'
+import { Vector3 } from 'three'
+import * as SkyboxFrag from './skybox.frag'
+import * as SkyboxVert from './skybox.vert'
```
BrendanAnnable: 👌

src/client/components/localisation/skybox/skybox.frag
```diff
@@ -0,0 +1,99 @@
+varying vec3 vWorldPosition;
```
BrendanAnnable: Where did these shaders come from?

src/client/components/localisation/skybox/view_model.ts
```diff
@@ -0,0 +1,80 @@
...
+export class SkyboxViewModel {
+  public constructor(private model: SkyboxModel) {
+  }
+
+  public static of = createTransformer((model: SkyboxModel): SkyboxViewModel => {
+    return new SkyboxViewModel(model)
+  })
+
+  @computed
+  public get skybox(): SkyBoxProps {
```
BrendanAnnable: It would be ideal to separate this function into multiple `@computed` properties which represented each `Mesh`/`Material`/`Geometry`. Not only just to keep functions small, but also for performance reasons. Mobx will cache each `@computed` property if no model changes occur that would require its re-computation. This is why I keep the geometries in the own `@computed` property since they are expensive to create and are otherwise completely static.

e.g. if the `luminance ` uniform were to change, we don't need to regenerate the ground, since it doesn't depend on the `luminance` uniform property. With it all being in one function like this, it would need to recompute all of it, when really only the sky changes.

Maybe use https://github.com/NUbots/NUsight2/blob/8715706c7e5b5034c04058bbb2eef027d9c2a531/src/client/components/localisation/field/view_model.ts#L21 as a good reference.

src/client/components/localisation/skybox/view_model.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  public static of = createTransformer((model: SkyboxModel): SkyboxViewModel => {
+    return new SkyboxViewModel(model)
+  })
+
+  @computed
+  public get skybox() {
+    // reference: http://threejs.org/examples/#webgl_shaders_sky
+    const shader = new SkyShader()
+    const uniforms = UniformsUtils.clone(shader.uniforms)
```
BrendanAnnable: Shouldn't the uniforms values come from the model? I think we can delete the `SkyShader` class and just inline the values directly here.

e.g.

```
const mat = new ShaderMaterial({
  fragmentShader: SkyboxFrag,
  vertexShader: SkyboxVert,
  uniforms: {
    luminance: { value: this.model.luminance },
    etc...
  },
  side: BackSide,
})
```

src/client/components/localisation/view_model.ts
```diff
@@ -21,7 +22,11 @@ export class LocalisationViewModel {
   public get scene(): Scene {
     const scene = new Scene()
     this.robots.forEach(robot => scene.add(robot))
+
     scene.add(this.field)
+    scene.add(this.skybox)
+    scene.add(this.skyboxGround)
+    scene.add(this.skyboxSun)
```
BrendanAnnable: I think you should combine these 3 into a single `this.skybox`, so that adding the skybox is a single scene addition e.g.

Usage: `scene.add(this.skybox)`

Impl:
```
@computed
private get skybox() {
   const skybox = new Object3D()
   skybox.add(this.skyboxSky) // <-- this is currently called skybox
   skybox.add(this.skyboxGround)
   skybox.add(this.skyboxSun)
   return skybox
}
```

src/client/components/localisation/skybox/view_model.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  @computed
+  public get sun() {
+    const sunSphere = new Mesh(
+      new SphereBufferGeometry( 40, 16, 8 ),
+      new MeshBasicMaterial( { color: 0xffffff } ),
+    )
+
+    sunSphere.position.y = - 49
```
BrendanAnnable: I don't think this line is used, because 3 lines down we call `sunSphere.position.copy`

src/client/components/localisation/skybox/view_model.ts
```diff
@@ -0,0 +1,88 @@
...
+    sunSphere.visible = false
```
BrendanAnnable: not used

src/client/components/localisation/skybox/view_model.ts
```diff
@@ -0,0 +1,88 @@
...
+    sunSphere.name = 'skyboxSun'
+    sunSphere.position.copy(this.sunPosition)
+    sunSphere.visible = this.model.sun
```
BrendanAnnable: ah, instead of setting the visibility, we should just not add it to the scene if `model.sun` is false

i.e. the `skybox.add(this.skyboxSun)` call should be wrapped in an if:

```
if (this.model.sun) {
  skybox.add(this.skyboxSun)
}
```

in order to make this easier, I think we should move the 3 skybox properties directly into `SkyboxViewModel` so it can control this.

e.g. In `localisation/view_model`, we only have a single prop:

```
@computed
private get skybox() {
  return SkyboxViewModel.of(this.model.skybox).skybox
}
```

and then in `skybox/view_model`:

```
@computed
private get skybox() {
  const skybox = new Object3D()
  skybox.add(this.sky)
  skybox.add(this.ground)
  if (this.model.sun) {
    skybox.add(this.sun)
  }
  return skybox
}

@computed
private get sky() {
  ...
}
@computed
private get ground() {
  ...
}
@computed
private get sun() {
  ...
}
```

src/client/components/localisation/skybox/model.ts
```diff
@@ -0,0 +1,48 @@
...
+  @observable public rayleigh: number
+  @observable public mieCoefficient: number
+  @observable public mieDirectionalG: number // default 0.8
+  @observable public luminance: number
+  @observable public inclination: number // elevation / inclination
+  @observable public azimuth: number // Facing front,
+  @observable public sun: boolean
+
+  public constructor(opts: SkyboxModelOpts) {
+    this.turbidity = opts.turbidity
```
BrendanAnnable: You can replace all these with `Object.assign(this, opts)`

src/client/components/localisation/skybox/model.ts
```diff
@@ -0,0 +1,48 @@
...
+export class SkyboxModel {
+
+  @observable public turbidity: number
+  @observable public rayleigh: number
+  @observable public mieCoefficient: number
+  @observable public mieDirectionalG: number // default 0.8
+  @observable public luminance: number
+  @observable public inclination: number // elevation / inclination
+  @observable public azimuth: number // Facing front,
+  @observable public sun: boolean
```
BrendanAnnable: Lets rename this to `showSun` while we're here

src/client/components/localisation/skybox/model.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  public constructor(opts: SkyboxModelOpts) {
```
BrendanAnnable: You can delete the `SkyboxModelOpts` interface and use itself as a type.

e.g. `public constructor(opts: SkyboxModel) {`

(I only use this convenient shorthand for pure data models such as this, though)

BrendanAnnable:
Might want to run the IntelliJ autoformatter on the skybox folder too. Resulted in a bunch of minor nit fixes 🙂 

screen shot 2017-06-19 at 1 14 38 am

src/client/components/localisation/skybox/view_model.ts
Phasedd: I was originally going to put it in `skybox/view_model`, but I got the impression it was to be in the `localisation/view_model` from the previous comment.

src/client/components/localisation/skybox/view_model.ts
```diff
@@ -0,0 +1,101 @@
...
+  @computed
+  private get sun() {
+    const sunSphere = new Mesh(
+      new SphereBufferGeometry(40, 16, 8),
+      new MeshBasicMaterial({ color: 0xffffff }),
+    )
+
+    sunSphere.name = 'skyboxSun'
+    sunSphere.position.copy(this.sunPosition)
+    sunSphere.visible = this.model.showSun
```
BrendanAnnable: You can delete this now 🙂

BrendanAnnable:
I dismissed your review @TrentHouliston — sorry 😛

Phasedd:
I was just about to dismiss his review too lol

PR #51, opened 2017-06-19
Cache yarn to speed up the build in travis

TrentHouliston:
By letting travis cache yarn dependencies we can speed up the build in travis a little.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,10 +6,14 @@ node_js:
   - '6'
   - '7'
   - '8'
+
+# Cache yarn and node_modules to speed up the build
 cache:
   yarn: true
   directories:
     - node_modules
+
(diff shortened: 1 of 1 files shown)


BrendanAnnable:
Did it make it faster?

TrentHouliston:
It's hard to tell, I see much smaller numbers for the yarn install stage but the total time isn't that different. But given how much travis varies depending on how it's feeling it's hard to tell overall if it's faster. But I mean it should be?

TrentHouliston:
Just for comparison, on the master branch's last build the yarn install step took 40.43s. But on the cached branch it took 1.67s. So no idea how the times end up so similar. I guess travis can vary by more than 30 seconds anyway.

PR #60, opened 2017-06-23
Upgrade to react router 4

BrendanAnnable:
- Upgrade to react router 4
- Also removed `mobx-react-router` since we're not using it.


src/client/components/localisation/view.tsx
```diff
@@ -59,7 +59,9 @@ export class LocalisationView extends React.Component<LocalisationViewProps, voi
         <LocalisationMenuBar onHawkEyeClick={this.onHawkEyeClick}/>
         <div className={style.localisation__canvasContainer}>
           <canvas className={style.localisation__canvas} ref={canvas => {
-            this.canvas = canvas
+            if (canvas) {
+              this.canvas = canvas
```
Phasedd: any particular reason why we are putting this in an if statement?

If canvas happens to be null / undefined, things are going to break anyway because this.canvas is going to be null / undefined.

BrendanAnnable: Added a handy `Preconditions.checkExists` helper which throws if `null/undefined`

BrendanAnnable:
Nice catch!

BrendanAnnable:
Discussed on slack, reverted to using `if` :slightly_smiling_face:

PR #66, opened 2017-07-05
Add robot picker [WIP]

JosephusPaye:
This adds a single robot picker like we had in Nusight.


JosephusPaye:
Closed in favour of what Brendan and Monica are working on.

PR #67, opened 2017-07-06
Robot selector

MonicaOlejniczak:
yay

4k glory screenshot

image


src/client/components/robot_selector/style.css
```diff
@@ -0,0 +1,62 @@
...
+  display: flex;
+  padding: 1em;
+}
+
+.robot:hover {
+  background-color: #eee;
+  cursor: pointer;
+}
+
+.robot:last-child:hover {
```
BrendanAnnable: If you add `overflow: hidden` to the `. robots` class you can avoid/delete this. A bit less coupling between the classes.

src/client/components/dropdown_container/view.tsx
```diff
@@ -0,0 +1,97 @@
+import { action } from 'mobx'
+import { observable } from 'mobx'
+import { observer } from 'mobx-react'
+import * as React from 'react'
+import { ComponentType } from 'react'
+
+export type WrappedComponentProps = {
```
BrendanAnnable: This is currently unused

src/client/components/dropdown_container/view.tsx
```diff
@@ -0,0 +1,97 @@
...
+  onToggle: () => void
+}
+
+export const dropdownContainer = (WrappedComponent: ComponentType<any>) => {
```
BrendanAnnable: I am not that familiar with HOC in react yet, so this could be standard practice (I think we briefly talked about it). But maybe we could simplify things by just tying `EnhancedDropdown` directly to `Dropdown`? Can you think of a likely scenario where we need to take advantage of the `WrappedComponent` parameter? A bit worried about over-engineering it.

src/client/components/switch/view.tsx
```diff
@@ -0,0 +1,34 @@
+import * as classNames from 'classnames'
+import * as React from 'react'
+import { ChangeEvent } from 'react'
+import { StatelessComponent } from 'react'
+import * as style from './style.css'
+
+export interface SwitchProps {
+  on: boolean
+  disabled?: boolean
+  onChange: (event: ChangeEvent<HTMLInputElement>) => void
```
BrendanAnnable: nit: `onChange(event: ChangeEvent<HTMLInputElement>): void`

src/client/components/robot_selector/view.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { Dropdown } from '../dropdown/view'
+import { dropdownContainer } from '../dropdown_container/view'
+import { Switch } from '../switch/view'
+import { RobotModel } from '../robot/model'
+import Robot from './robot.svg'
+import * as style from './style.css'
+
+export type RobotSelectorProps = {
+  robots: RobotModel[]
+  selectRobot: (robot: RobotModel) => void
```
BrendanAnnable: nit: `selectRobot(robot: RobotModel): void`

src/client/components/app/controller.ts
```diff
@@ -0,0 +1,15 @@
+import { action } from 'mobx'
+import { RobotModel } from '../robot/model'
+
+export class AppController {
+
```
MonicaOlejniczak: Shouldn't this be a lint rule then?

src/client/components/app/controller.ts
```diff
@@ -0,0 +1,15 @@
...
+  public static of() {
+    return new AppController()
+  }
+
+  @action
+  public toggleRobot(model: RobotModel) {
```
MonicaOlejniczak: Sure

src/client/components/app/controller.ts
BrendanAnnable: That would be good, are you aware of one that does this?

src/client/components/menu_bar/style.css
```diff
@@ -2,3 +2,11 @@
   display: flex;
   height: 60px;
 }
+
+.menu {
+  flex: 1
+}
+
+.robotSelector {
```
MonicaOlejniczak: Was for debugging purposes before, it can be useful.

src/client/components/robot_selector/view.tsx
MonicaOlejniczak: Also should be a lint rule if possible?

src/client/components/robot_selector/view.tsx
```diff
@@ -0,0 +1,43 @@
...
+export const RobotSelector = observer((props: RobotSelectorProps) => {
+  const { robots, selectRobot } = props
+  const dropdownToggle = (
+    <button className={style.button}>
+      <Robot/>
+      Select robots
+    </button>
+  )
+  const onChange = (robot: RobotModel) => () => selectRobot(robot)
+  const dropdownMenuStyle = { right: 0 }
```
MonicaOlejniczak: Nope

src/client/components/dropdown_container/view.tsx
MonicaOlejniczak: Ya, was leftover from many refactors.

src/client/components/app/controller.ts
BrendanAnnable: Looks like we want http://eslint.org/docs/rules/padded-blocks

Unfortunately `tslint` nor `tslint-eslint-rules` supports it 😞

src/client/components/dropdown_container/view.tsx
```diff
@@ -0,0 +1,97 @@
...
+        this.isOpen = false
+      }
+    }
+
+    private onDocumentClick(event: MouseEvent) {
+      if (!this.dropdown) {
+        return
+      }
+
+      const isOutsideEl = (target: EventTarget, el?: HTMLElement): boolean => {
```
MonicaOlejniczak: Yup

src/client/components/robot_selector/view.tsx
BrendanAnnable: Looks like it https://palantir.github.io/tslint/rules/prefer-method-signature/

src/client/components/robot_selector/style.css
MonicaOlejniczak: Yep, that's definitely better. I did use that for dashboard related things but it didn't come to mind in this context.

BrendanAnnable:
Please add an empty state for when no robots exist 🙂

PR #69, opened 2017-07-10
Add improved networking

BrendanAnnable:
This PR is mostly about refactoring the network-level architecture such that it is testable, more reliable and easy to use. Component-level code remains unchanged.

Features:
- [x] NUClearNet abstractions!
All NUClearNet usage hides behind a new `NUClearNetClient` interface, which has 3 separate implementations:
  - [x] `DirectNUClearNetClient` — An adapter to a real NUClearNet connection, over a real socket.
  - [x] `FakeNUClearNetClient` — A fake connection, no sockets. Used for local development (with `yarn start:sim`) and also for testing. Collaborates with `FakeNUClearNetServer`.
  - [x] `WebSocketProxyNUClearNetClient` — A magical client-side proxy which lets us pretend we have a `NUClearNet` connection directly from within the browser. All requests get proxied over to `WebSocketProxyNUClearNetServer`, which will either forward them to a real NUClearNet connection or a fake one, depending on whether you used `yarn start` or `yarn start:sim`. This has simplified the networking internals as now I can keep all `NUClearNet` traffic on a separate socket.io namespace and keep it separate from other NUsight-to-node communications (recording for instance). It also re-establishes listeners on any disconnection, such that server crashes/restarts should never need a browser reload!
- [x] Super high level robot-to-component network integration tests.
- [x] Unit tests for a bunch-of-stuff.
- [x] Deleted currently unused things such as the `NUSightServer` class — it will make a come back shortly.
- [x] Renamed `GlobalNetwork` to `NUsightNetwork`
- [x] Created `createSingletonFactory` to help with singleton factories.

Next PR will add robot sync, which will hopefully be easier now.


src/server/nuclearnet/tests/fake_nuclearnet_client.tests.ts
```diff
@@ -0,0 +1,192 @@
...
+    eve.send({ type: 'sensors', payload })
+
+    expect(bobOnSensors).toHaveBeenCalledTimes(1)
+    expect(bobOnSensors).toHaveBeenLastCalledWith({ payload, peer: expect.objectContaining({ name: 'eve' }) })
+
+    expect(aliceOnSensors).toHaveBeenCalledTimes(1)
+    expect(aliceOnSensors).toHaveBeenLastCalledWith({ payload, peer: expect.objectContaining({ name: 'eve' }) })
+
+    expect(eveOnSensors).toHaveBeenCalledTimes(1)
+    expect(aliceOnSensors).toHaveBeenLastCalledWith({ payload, peer: expect.objectContaining({ name: 'eve' }) })
```
TrentHouliston: should be eveOnSensors

src/server/nuclearnet/tests/fake_nuclearnet_client.tests.ts
```diff
@@ -0,0 +1,192 @@
...
+  })
+
+  it('only receives targetted messages if they are the target', () => {
```
TrentHouliston: You should also test that two clients with the same name will both receive messages sent to a target

src/server/nuclearnet/tests/fake_nuclearnet_client.tests.ts
```diff
@@ -0,0 +1,192 @@
+import { FakeNUClearNetClient } from '../fake_nuclearnet_client'
+import { FakeNUClearNetServer } from '../fake_nuclearnet_server'
+
+describe('FakeNUClearNetClient', () => {
```
TrentHouliston: If you really want to ensure that you're simulating NUClear properly there are a few other things you should put in the tests/code as they may have an influence on the behaviour of the system.
- when connecting a single client will connect from an address/port. This port is assigned by the OS and if you disconnect and reconnect, you'll often have a different port.
- the address may be either an ipv4 or ipv6 address, so make sure you don't assume it's a dotted decimal address.

src/client/network/network.ts
```diff
@@ -1,37 +1,31 @@
...
-import { MessageType } from './global_network'
-import { Message } from './global_network'
-import { MessageCallback } from './global_network'
+import { NUsightNetwork } from './nusight_network'
+import { MessageType } from './nusight_network'
+import { MessageCallback } from './nusight_network'
 
 export class Network {
-  private listeners: Map<MessageType<Message>, Set<MessageCallback<Message>>>
+  private offs: (() => void)[]
```
BrendanAnnable: Done.

src/client/network/nusight_network.ts
```diff
@@ -0,0 +1,42 @@
+import { NUClearNetPacket } from 'nuclearnet.js'
+import { NUClearNetOptions } from 'nuclearnet.js'
+import { NUClearNetClient } from '../../shared/nuclearnet/nuclearnet_client'
+import { message } from '../../shared/proto/messages'
```
BrendanAnnable: Removed.

src/client/network/nusight_network.ts
```diff
@@ -0,0 +1,42 @@
...
+import { WebSocketProxyNUClearNetClient } from '../nuclearnet/web_socket_proxy_nuclearnet_client'
+import { MessageTypePath } from './message_type_names'
+import Sensors = message.input.Sensors
```
BrendanAnnable: Removed.

src/client/nuclearnet/web_socket_proxy_nuclearnet_client.ts
```diff
@@ -0,0 +1,108 @@
...
+      this.socket.send(options.type, options)
+    }
+  }
+
+  private onReconnect = (options: NUClearNetOptions) => {
+    this.socket.send('nuclear_connect', options)
+
+    // Restore all our listeners
+    for (const joinListener of this.joinListeners) {
+      this.socket.on('nuclear_join', joinListener)
```
BrendanAnnable: Added a comment. The server has no persistence, so if it crashes or restarts, there is no way it knows a client was previously listening to. So as a client we have to subscribe to all those events again.

src/server/nuclearnet/fake_nuclearnet_client.ts
```diff
@@ -0,0 +1,90 @@
...
+   * between tests. Simply use the constructor of both FakeNUClearNetServer and FakeNUClearNetClient instead.
+   */
+  public static of(): FakeNUClearNetClient {
+    return new FakeNUClearNetClient(FakeNUClearNetServer.of())
+  }
+
+  public connect(options: NUClearNetOptions): () => void {
+    this.peer = {
+      name: options.name,
+      address: '127.0.0.1',
```
BrendanAnnable: Those values aren't used, I've updated them to be much more fake.

src/server/nuclearnet/fake_nuclearnet_server.ts
```diff
@@ -0,0 +1,64 @@
...
+    this.events = new EventEmitter()
+    this.clients = []
+  }
+
+  /**
+   * Designed to only have a single instance during runtime. This singleton is mimicking the behaviour of
+   * the real NUClearNet, where a single global TCP/IP network is an implicit non-injectable dependency. This is why
+   * FakeNUClearNetClient.of() does not take a given server and instead uses the singleton instance.
+   *
+   * Avoid using this singleton factory in tests though, as you'll have introduce cross-contamination between tests.
```
BrendanAnnable: Fixed.

src/server/nuclearnet/tests/fake_nuclearnet_client.tests.ts
BrendanAnnable: I don't use address/port anywhere in the fake, so hopefully I can ignore those things? I've updated the dummy values to be very fake such that nobody would be able to think they work.

src/server/nuclearnet/tests/fake_nuclearnet_client.tests.ts
BrendanAnnable: oo nice catch. Fixed!

src/server/nuclearnet/tests/fake_nuclearnet_client.tests.ts
BrendanAnnable: Added.

src/server/nuclearnet/fake_nuclearnet_server.ts
```diff
@@ -0,0 +1,64 @@
...
+   */
+  public static of = createSingletonFactory(() => {
+    return new FakeNUClearNetServer()
+  })
+
+  public connect(client: FakeNUClearNetClient): () => void {
+    this.events.emit('nuclear_join', client.peer)
+    this.clients.push(client)
+
+    for (const otherClient of this.clients) {
```
BrendanAnnable: Added.

BrendanAnnable:
@TrentHouliston changes in!

src/client/nuclearnet/web_socket_proxy_nuclearnet_client.ts
```diff
@@ -6,14 +6,21 @@ import { NUClearNetClient } from '../../shared/nuclearnet/nuclearnet_client'
...
+
+  // Store all listeners so that we can restore them after a socket disconnection/reconnection.
   private joinListeners: Set<NUClearEventListener>
   private leaveListeners: Set<NUClearEventListener>
-  private packetListeners: Map<string, Set<{ messageId: string, listener: NUClearPacketListener }>>
+  private packetListeners: Map<string, Set<{ requestToken: string, listener: NUClearPacketListener }>>
 
   public constructor(private socket: WebSocketClient) {
-    this.nextMessageId = 0
+    this.nextRequestToken = 0
```
TrentHouliston: You might be able to replace this functionality by using socket.ios ability to send callbacks to the server and having it execute them there. Or sending callbacks from the server to the client, either way works.

src/client/nuclearnet/web_socket_proxy_nuclearnet_client.ts
TrentHouliston: Would this cause a problem if the server didn't crash? Does the server purge our connections when we disconnect? Otherwise we may end up double subscribing to messages.

src/server/nuclearnet/fake_nuclearnet_client.ts
```diff
@@ -29,8 +36,9 @@ export class FakeNUClearNetClient implements NUClearNetClient {
   public connect(options: NUClearNetOptions): () => void {
     this.peer = {
       name: options.name,
-      address: '127.0.0.1',
-      port: 7447,
+      // Address and port are not used for anything yet, dummy values used.
+      address: 'fake_address',
```
TrentHouliston: Make sure you don't try to decode this anywhere as an IP address :P

BrendanAnnable: haha indeed! The idea is that any code that tries to will break since its not designed to be used.

src/server/nuclearnet/fake_nuclearnet_client.ts
```diff
@@ -29,8 +36,9 @@ export class FakeNUClearNetClient implements NUClearNetClient {
...
+      port: 0,
```
TrentHouliston: A port of -1 is probably more valid as an invalid port. Since "technically" port 0 is a valid port.

BrendanAnnable: `Infinity`? 😛

TrentHouliston: any number thats 0 > x > 65535 is invalid enough :P

BrendanAnnable: I'll use 65535.1

TrentHouliston: But what if you accidentally cast it to an int! Then it would drop the 0.1, clearly it needs to be 65536.1

BrendanAnnable: Exactly! 😛  What if -1 gets cast to a Uint16 and wraps to 65535! 😆 

I think I'm actually going to use `NaN` lol

TrentHouliston: Well any Uint16 is valid so if it gets in one of those it's all over :P 
Wonder what happens when you cast NaN to a unit16

TrentHouliston: I was thinking more
untitled 2

src/client/nuclearnet/web_socket_proxy_nuclearnet_client.ts
BrendanAnnable: The problem with this is that the unsubscribe function (i.e. the `off()`) must be returned from `on()` synchronously. Getting an off method from the server will be async and can't be returned 😞 .

BrendanAnnable:
@TrentHouliston changes in!

PR #74, opened 2017-07-14
Upgrade simulator to simulate multiple robots

BrendanAnnable:
This change will allow `yarn start:sim` to simulate more than a single robot, which is great for testing a larger range of scenarios, especially performance tuning. `SimulatorStatus` got cut (the thing that printed simulated messages/second to the console), but I'll revisit it later.

Does not include network throttling code, so I'm purposefully keeping it at 1 for now so that your browser does not crash 🙂 

I'll increase the default `numRobots` count when a few more related PRs go through.


src/simulators/simulator.ts
```diff
@@ -1,5 +1,5 @@
 export interface Simulator {
-  simulate(time: number): Message[]
+  simulate(time: number, index: number, numRobots: number): Message[]
```
BrendanAnnable: Adding those 2 params allow the simulators to act uniquely per robot.

PR #75, opened 2017-07-14
Add support for Sensors world transform

BrendanAnnable:
Removes fake client-side robot movement in favour of interpreting the world transform on the `message.input.Sensors` message.

Also updates `SensorDataSimulator` to match.


src/client/components/localisation/network.ts
```diff
@@ -22,27 +27,49 @@ export class LocalisationNetwork {
...
-    robot.motors.rightAnklePitch.angle = Number(sensors.servo[14].presentPosition)
-    robot.motors.leftAnklePitch.angle = Number(sensors.servo[15].presentPosition)
-    robot.motors.rightAnkleRoll.angle = Number(sensors.servo[16].presentPosition)
-    robot.motors.leftAnkleRoll.angle = Number(sensors.servo[17].presentPosition)
-    robot.motors.headPan.angle = Number(sensors.servo[18].presentPosition)
-    robot.motors.headTilt.angle = Number(sensors.servo[19].presentPosition)
+
+    const { translation, rotation } = decompose(fromProtoMat44(sensors.world!))
+    robot.position.set(translation.x, translation.z, translation.y)
+    robot.heading = new Euler().setFromQuaternion(rotation).z
```
TrentHouliston: Should put the whole quaternion on the model as it's all used. Also robot.heading could be confusing as localisation will likely want to use a similar name. Ideally we should start using the same naming convention we are using on the robots. So that would make this Rtw or Rwt depending on if it was transposed.

src/client/index.tsx
```diff
@@ -47,27 +47,22 @@ runInAction(() => {
...
+      color: colors[id] || undefined,
+      heading: -angle - Math.PI / 2
+    })
+    robot.position.x = distance * Math.cos(angle)
+    robot.position.z = distance * Math.sin(angle)
+
     localisationModel.robots.push(robot)
     return robot
   })
 })
```
BrendanAnnable: This entire block will soon be gone.

src/client/components/localisation/network.ts
BrendanAnnable: Done!

src/client/components/localisation/controller.ts
```diff
@@ -214,8 +216,10 @@ export class LocalisationController {
 
     // This camera position hack will not work with orientation/head movement.
     // TODO (Annable): Sync camera position/rotation properly using kinematic chain.
-    model.camera.position.set(target.position.x - 0.001, target.position.y + 0.4, target.position.z)
-    model.camera.yaw = target.heading + Math.PI // TODO (Annable): Find why offset by PI is needed.
+    model.camera.position.set(target.rWTt.x - 0.001, target.rWTt.y + 0.4, target.rWTt.z)
+    const Rwt = new Quaternion(target.Rwt.x, target.Rwt.y, target.Rwt.z, target.Rwt.w)
+    const heading = new Euler().setFromQuaternion(Rwt).y
+    model.camera.yaw = heading + Math.PI // TODO (Annable): Find why offset by PI is needed.
```
BrendanAnnable: Hm this is still wrong.

src/client/components/localisation/darwin_robot/model.ts
```diff
@@ -1,22 +1,24 @@
...
 import { Vector3 } from '../model'
+import { Quaternion } from '../model'
 
 export class RobotModel {
   @observable public id: number
   @observable public name: string
   @observable public color?: string
-  @observable public heading: number
-  @observable public position: Vector3
+  @observable public rWTt: Vector3 // Torso to world translation
```
TrentHouliston: Add in torso space to the comment might be good.

BrendanAnnable:
First person mode is broken, but it needs a large refactor to work properly with quaternions. I'm deferring that to a follow-up PR.

PR #76, opened 2017-07-14
Add networked robots

BrendanAnnable:
This PR adds full networking support for robots. That is to say, newly connected robots will automatically show up in the robot selector as well as in the localisation display.

Looks/works better with #75 + throttling though.

### DIFF
--- a/src/client/components/localisation/network.ts
+++ b/src/client/components/localisation/network.ts
@@ -2,8 +2,10 @@ import { action } from 'mobx'
 import { message } from '../../../shared/proto/messages'
 import { Network } from '../../network/network'
 import { NUsightNetwork } from '../../network/nusight_network'
+import { RobotModel } from '../robot/model'
 import { LocalisationModel } from './model'
 import Sensors = message.input.Sensors
+import { LocalisationRobotModel } from './darwin_robot/model'
 
 export class LocalisationNetwork {
   public constructor(private network: Network,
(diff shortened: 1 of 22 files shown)


src/client/components/localisation/network.ts
```diff
@@ -21,9 +23,8 @@ export class LocalisationNetwork {
   }
 
   @action
-  private onSensors = (sensors: Sensors) => {
-    // TODO (Annable): Remove hardcoded values.
-    const robot = this.model.robots[0]
+  private onSensors = (robotModel: RobotModel, sensors: Sensors) => {
+    const robot = LocalisationRobotModel.of(robotModel)
```
BrendanAnnable: yayyy! 🎉

src/server/nuclearnet/fake_nuclearnet_client.ts
```diff
@@ -36,9 +36,9 @@ export class FakeNUClearNetClient implements NUClearNetClient {
   public connect(options: NUClearNetOptions): () => void {
     this.peer = {
       name: options.name,
-      // Address and port are not used for anything yet, dummy values used.
-      address: 'fake_address',
-      port: NaN,
+      address: '127.0.0.1',
```
BrendanAnnable: I'm not sure what this value should be?

src/server/nuclearnet/fake_nuclearnet_client.ts
```diff
@@ -36,9 +36,9 @@ export class FakeNUClearNetClient implements NUClearNetClient {
...
+      // TODO (Annable): Inject a random function.
+      port: Math.floor(Math.random() * (65536 - 1024) + 1024),
```
TrentHouliston: Ephemeral ports are usually much larger than 1024... The Internet Assigned Numbers Authority (IANA) suggests the range 49152 to 65535 (2^15+2^14 to 2^16−1) for dynamic or private ports.
Bu this is probably fine.

PR #77, opened 2017-07-15
Add overview simulator

BrendanAnnable:
Direct port from NUsight1, minimal thought went into refactoring.

I've also just realised that a simulator no longer has control over its own frequency, so this would be sending 60 overview messages/second.

Probably needs some reworking. Handing this code and PR over to you @MonicaOlejniczak 🙂 Adapt as you need.


src/simulators/overview_simulator.ts
```diff
@@ -0,0 +1,84 @@
...
+      ballPosition,
+      ballWorldPosition,
+      gameMode: Mode.NORMAL,
+      gamePhase: Phase.INITIAL,
+      penaltyReason: PenaltyReason.UNPENALISED,
+      lastCameraImage: { seconds: time - 5000 * this.random() },
+      lastSeenBall: { seconds: time - 5000 * this.random() },
+      lastSeenGoal: { seconds: time - 15000 * this.random() },
+      lastSeenObstacle: { seconds: time - 15000 * this.random() },
+      pathPlan: [
```
BrendanAnnable: Is this no longer a thing @TrentHouliston?

src/simulators/overview_simulator.ts
```diff
@@ -0,0 +1,84 @@
...
+        robotPosition,
+        this.randomFieldPosition(),
+        this.randomFieldPosition(),
+        this.randomFieldPosition(),
+        ballWorldPosition
+      ],
+      kickTarget: this.randomFieldPosition(),
```
BrendanAnnable: Is this still a thing? @TrentHouliston

src/server/prod.ts
```diff
@@ -35,6 +38,7 @@ if (withSimulators) {
     numRobots: 1,
     simulators: [
       SensorDataSimulator.of(),
+      OverviewSimulator.of(),
     ],
   })
   virtualRobots.simulateWithFrequency(60)
```
BrendanAnnable: - [ ] TODO: I'll have to wrap each simulator separately.

src/simulators/overview_simulator.ts
```diff
@@ -0,0 +1,84 @@
...
+      robotHeading,
+      ballPosition,
+      ballWorldPosition,
+      gameMode: Mode.NORMAL,
+      gamePhase: Phase.INITIAL,
+      penaltyReason: PenaltyReason.UNPENALISED,
+      lastCameraImage: { seconds: time - 5000 * this.random() },
+      lastSeenBall: { seconds: time - 5000 * this.random() },
+      lastSeenGoal: { seconds: time - 15000 * this.random() },
+      lastSeenObstacle: { seconds: time - 15000 * this.random() },
```
BrendanAnnable: Do we detect obstacles?

TrentHouliston: We do not detect obstacles at the moment

src/simulators/overview_simulator.ts
TrentHouliston: The path plan doesn't exist in that form any more no. But we do have a ball approach position point that we could use.

src/simulators/overview_simulator.ts
TrentHouliston: We still have a kick target

PR #83, opened 2017-07-15
Add network throttling

BrendanAnnable:
- Throttles unreliable packets so that we do not overwhelm the client with traffic.
- Sits between `WebSocketProxyNUClearNetServer` and `WebSocketProxyNUClearNetClient`.
- The limit is per-event, such that no particular event should be starved more than any other.
- The limit is per-robot, such that no robot is starved more than any other.

### DIFF
--- a/src/server/nuclearnet/web_socket_proxy_nuclearnet_server.ts
+++ b/src/server/nuclearnet/web_socket_proxy_nuclearnet_server.ts
@@ -6,6 +6,8 @@ import { DirectNUClearNetClient } from './direct_nuclearnet_client'
 import { FakeNUClearNetClient } from './fake_nuclearnet_client'
 import { WebSocketServer } from './web_socket_server'
 import { WebSocket } from './web_socket_server'
+import { NodeSystemClock } from '../time/node_clock'
+import { Clock } from '../time/clock'
 
 type Opts = {
   fakeNetworking: boolean
@@ -36,12 +38,14 @@ class WebSocketServerClient {
   private offJoin: () => void
   private offLeave: () => void
   private offListenMap: Map<string, () => void>
+  private processors: Map<NUClearNetPeer, RobotMessageProcessor>
 
   public constructor(private nuclearnetClient: NUClearNetClient, private socket: WebSocket) {
     this.connected = false
     this.offJoin = this.nuclearnetClient.onJoin(this.onJoin)
     this.offLeave = this.nuclearnetClient.onLeave(this.onLeave)
     this.offListenMap = new Map()
+    this.processors = new Map()
 
     this.socket.on('listen', this.onListen)
     this.socket.on('unlisten', this.onUnlisten)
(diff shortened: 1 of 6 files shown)


src/server/nuclearnet/web_socket_proxy_nuclearnet_server.ts
```diff
@@ -93,6 +102,83 @@ class WebSocketServerClient {
...
+  constructor(private socket: WebSocket,
+              private clock: Clock,
+              opts: { limit: number, timeout: number }) {
+    this.limit = opts.limit
+    this.timeout = opts.timeout
+    this.eventQueueSize = new Map()
+  }
+
+  public static of(socket: WebSocket) {
+    return new RobotMessageProcessor(socket, NodeSystemClock, { limit: 1, timeout: 5000 })
```
BrendanAnnable: Increasing `limit > 1` seems to have significant slowdown. I suspect a web worker pool would let us increase this 🙂

src/server/nuclearnet/web_socket_proxy_nuclearnet_server.ts
```diff
@@ -93,6 +102,83 @@ class WebSocketServerClient {
...
+    if (peerKey) {
+      const processor = this.processors.get(peerKey)
+      if (processor) {
+        processor.onPacket(event, packet)
+      }
+    }
+  }
+
+  /**
+   * Look for the needle 'peer' inside the haystack 'peers'. They might not be the same object reference.
```
TrentHouliston: I'm sorry!

TrentHouliston: also, not might not... will not

src/server/nuclearnet/web_socket_proxy_nuclearnet_server.ts
```diff
@@ -93,6 +102,83 @@ class WebSocketServerClient {
...
+      return otherPeer.name === peer.name && otherPeer.address === peer.address && otherPeer.port === peer.port
+    })
+  }
+}
+
+class RobotMessageProcessor {
+  private eventQueueSize: Map<string, number>
+
+  // The maximum number of packets of a unique type to send before receiving acknowledgements.
+  private limit: number
```
TrentHouliston: I wonder if you can use tcp's slow start scheme here to find the limit dynamically. Although the advantage to that would be latency masking rather than increase in bandwidth. Could be useful for situations like the heroku app.

src/server/nuclearnet/web_socket_proxy_nuclearnet_server.ts
BrendanAnnable: how could you! 😭 😛 

haha well, _will not_ in the current use case, but this method could take any list of peers and any peer so I wanted to remain ambiguous.

src/server/nuclearnet/web_socket_proxy_nuclearnet_server.ts
BrendanAnnable: That does sound useful!

PR #87, opened 2017-07-15
Use webworkers to decode protocol buffers from the network

TrentHouliston:
A significant amount of processing is done to decode messages from the network.
We can offload this to other threads to keep the main browser responsive and improve performance.

### DIFF
--- a/src/client/network/decode.worker.ts
+++ b/src/client/network/decode.worker.ts
@@ -1,5 +1,22 @@
-import { MessageTypePath } from './message_type_names'
+import { message } from '../../shared/proto/messages'
+
+const HEADER_SIZE = 9
 
 addEventListener('message', (e: MessageEvent) => {
-  postMessage({ 'token': 'Hello main thread!!' })
+  const token = e.data.token
+  const type = e.data.type.split('.')
+  const payload = e.data.payload
+
+  // Using our message type access the appropriate decoder
+  let decoder: any = message
+  for (let i = 1; i < type.length; ++i) {
+    decoder = decoder[type[i]]
+  }
+
+  // Remove NUsight header for decoding by protobufjs
+  const buffer = new Uint8Array(payload).slice(HEADER_SIZE)
+  const output = decoder.decode(buffer)
+
+  // Send back our decoded data!
+  postMessage({ 'token': token, 'output': output })
 })
--- a/src/client/network/decode.worker.ts
+++ b/src/client/network/decode.worker.ts
@@ -1,5 +1,5 @@
 import { MessageTypePath } from './message_type_names'
 
-self.addEventListener('message', (e: MessageEvent) => {
-  self.postMessage({ 'token': 'Hello main thread!!' })
+addEventListener('message', (e: MessageEvent) => {
+  postMessage({ 'token': 'Hello main thread!!' })
 })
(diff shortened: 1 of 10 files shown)


src/client/network/nusight_network.ts
```diff
@@ -33,16 +87,16 @@ export class NUsightNetwork {
   public onNUClearMessage<T>(messageType: MessageType<T>, cb: MessageCallback<T>) {
     const messageTypeName = this.messageTypePath.getPath(messageType)
     return this.nuclearnetClient.on(`NUsight<${messageTypeName}>`, (packet: NUClearNetPacket) => {
-      // Remove NUsight header for decoding by protobufjs
-      const buffer = new Uint8Array(packet.payload).slice(HEADER_SIZE)
-      const message = messageType.decode(buffer)
```
BrendanAnnable: I have no idea how, but you'll need to pass `messageType.decode` to the worker somehow, because the decoding process is different per message... 😞

TrentHouliston: I'm just going to send the message type as a string and let it work it out itself using the message_type_names.ts

src/client/network/nusight_network.ts
```diff
@@ -6,8 +6,69 @@ import { WebSocketProxyNUClearNetClient } from '../nuclearnet/web_socket_proxy_n
...
+  }
+
+  private onTaskComplete(worker: Worker, msg: {token: string, output: any}) {
+
+    // Find and execute our callback
+    this.callbacks[msg.token](msg.output)
+    delete this.callbacks[msg.token]
+
+    // If this worker has finished all its tasks move it to the free list
+    const w = this.busy.splice(this.busy.indexOf(worker), 1)[0]
```
TrentHouliston: This is not ideal, i'd like to only remove it from busy if it actually has no other task it's running

src/client/network/decode.worker.ts
```diff
@@ -0,0 +1,22 @@
...
+  for (let i = 1; i < type.length; ++i) {
+    decoder = decoder[type[i]]
+  }
+
+  // Remove NUsight header for decoding by protobufjs
+  const buffer = new Uint8Array(payload).slice(HEADER_SIZE)
+  const output = decoder.decode(buffer)
+
+  // Send back our decoded data!
+  postMessage({ 'token': token, 'output': output })
```
TrentHouliston: It seems this can sometimes fail and cause an error which stops this worker getting removed from busy.

src/client/network/nusight_network.ts
```diff
@@ -6,8 +6,69 @@ import { WebSocketProxyNUClearNetClient } from '../nuclearnet/web_socket_proxy_n
...
+    worker = worker === undefined && packet.reliable ? this.busy.shift() : worker
+
+    // We drop packets if they are unreliable and we are too busy to process them
+    if (worker !== undefined) {
+      // Store our callback with a unique token so we can find it later
+      const token = ++this.tokenSource
+      this.callbacks[token] = cb
+
+      // Execute on the webworker
+      worker.postMessage({
```
TrentHouliston: Sometimes this can cause an error meaning the worker is lost.

ysims:
Old and unlikely to be finished. Can be recoded in future if needed.

PR #92, opened 2017-07-20
Update all timers to use seconds

BrendanAnnable:
In the interest of keeping all our units consistently using the SI standard and reducing the number of conversions throughout the code, this PR updates all timers to operate in seconds.


BrendanAnnable:
It could, but once they do know what's happening things _should_ be easier.

PR #97, opened 2017-07-20
Add NBS streaming primitives

BrendanAnnable:
This PR adds the building blocks for adding recording and playback. The frontend UI and integration with these classes will come later.

Firstly there is the core `encodeFrame` and `decodeFrame` functions. These encode and decode NBS frames/buffers. They are raw simply raw functions which pack/unpack the data to/from a binary format.

The rest of the classes are sugar designed to make it easy to use these functions in practical ways. They are written using node streams, which allow us to easily pipe data around, such as piping a gzipped `.nbz` file straight onto the NUClear network.

Overall this is how it works:

`FileStream` -> `NbsFrameChunker` -> `NbsFrameDecoder` -> `NbsNUClearPlayback` -> `NUClearNetClient` -> `Browser and/or Robot`.

`NbsFrameChunker` — The entire job of this stream tranformer is to find the nbs header (0xe298a2) and read just enough to know its size, and then form nbs-packet sized buffer chunks. In other words it decodes the framing protocol and emits those frames in the form of buffers.

`NbsFrameDecoder` — This is the next step, actually decoding that framed buffer into a JavaScript object with readable properties. There is also its partner `NbsFrameDecoder`, which should let us stream those same JavaScript objects to a file.

`NbsNUClearPlayback` — The job of this class is to take the JavaScript frames and put them on the network. It also streams them at the same rate they were recorded. i.e. it reads the `timestampInMicroseconds` property and emits packets at the appropriate time to simulate playback.

Now just look at those node streams in action 😍 :

```
  public static fromFile(filename: string, nuclearnetClient: NUClearNetClient) {
    const playback = NbsNUClearPlayback.of(nuclearnetClient)
    let rawStream = fs.createReadStream(filename)
    const decompress = filename.endsWith('.nbz') ? createGunzip() : new PassThrough()
    rawStream.pipe(decompress).pipe(new NbsFrameChunker()).pipe(new NbsFrameDecoder()).pipe(playback)
    return playback
  }
```


src/server/nbs/nbs_frame_codecs.ts
```diff
@@ -0,0 +1,49 @@
...
+// 3 Bytes - NUClear radiation symbol header, useful for synchronisation when attaching to an existing stream.
+// 4 Bytes - The remaining packet length i.e. 16 bytes + N payload bytes
+// 8 Bytes - 64bit timestamp in microseconds. Note: this is not necessarily a unix timestamp.
+// 8 Bytes - 64bit bit hash of the message type.
+// N bytes - The binary packet payload.
+
+export function encodeFrame(frame: NbsFrame): Buffer {
+  const size = 16 + frame.payload.byteLength
+  const buffer = new Buffer(7 + size)
+  NBS_HEADER.copy(buffer, 0, 0, 3)
```
TrentHouliston: Nodejs has a function Buffer.concat([buffer1,buffer2...]) function. Having each component as it's own buffer and concatenating at the end might remove some of these magic numbers.

src/server/nbs/nbs_frame_chunker.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    done()
+  }
+
+  private getNextFrame(buffer: Buffers): { offset: number, buffer: Buffer } | undefined {
+    const headerIndex = buffer.indexOf(NBS_HEADER)
+    const headerSize = NBS_HEADER.byteLength
+    const packetLengthSize = 4
+    const headerAndPacketLengthSize = headerSize + packetLengthSize
+    if (headerIndex >= 0) {
```
TrentHouliston: The block here/below is very dense and could probably use some comments

src/server/nbs/nbs_nuclear_playback.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  public static of(nuclearnetClient: NUClearNetClient) {
+    return new NbsNUClearPlayback(nuclearnetClient, NodeSystemClock)
+  }
+
+  public static fromFile(filename: string, nuclearnetClient: NUClearNetClient) {
+    const playback = NbsNUClearPlayback.of(nuclearnetClient)
+    let rawStream = fs.createReadStream(filename)
+    const decompress = filename.endsWith('.nbz') ? createGunzip() : new PassThrough()
```
TrentHouliston: This should be ends with .nbz or .nbs.gz both are valid extensions

src/server/nbs/tests/nbs_nuclear_playback.tests.ts
```diff
@@ -39,15 +39,15 @@ describe('NbsNUClearPlayback', () => {
     expect(nuclearnetClient.send).toHaveBeenCalledWith(expect.objectContaining({ type: hashType('frame1') }))
 
     playback.write(frame2)
-    clock.runTimersToTime(1009) // Should have sent packet yet at only 9 elapsed seconds
+    clock.runTimersToTime(1009) // Should not have sent packet yet at only 9 elapsed seconds.
     expect(nuclearnetClient.send).not.toHaveBeenCalledWith(expect.objectContaining({ type: hashType('frame2') }))
-    clock.runTimersToTime(1010) // At 10 elapsed seconds the packet should have been sent
+    clock.runTimersToTime(1010) // At 10 elapsed seconds the packet should have been sent.
```
TrentHouliston: Aren't these times supposed to be in seconds?

PR #102, opened 2017-07-23
Introduce groups and local transforms to dashboard

MonicaOlejniczak:
- Create scene and group objects.
- Update canvas renderer to use a scene.
- Update object2d type to include add method and transform.
- Update geometry type to use a transform instead of a object2d.


src/client/canvas/renderer.ts
```diff
@@ -43,40 +43,53 @@ export class CanvasRenderer {
...
+          y: 1 / obj.transform.scale.y,
+        },
+        translate: {
+          x: 0,
+          y: 0,
+        },
+      })
+      this.context.scale(obj.transform.scale.x, obj.transform.scale.y)
+      this.context.rotate(-obj.transform.rotate)
+      this.context.translate(translateDash.x, translateDash.y)
```
BrendanAnnable: Could we factor this above code out into a `applyTransform(context, transform)` function?

src/client/canvas/renderer.ts
```diff
@@ -43,40 +43,53 @@ export class CanvasRenderer {
...
+      this.renderCircle({ appearance, geometry })
+    } else if (geometry instanceof LineGeometry) {
+      this.renderLine({ appearance, geometry })
+    } else if (geometry instanceof MarkerGeometry) {
+      this.renderMarker({ appearance, geometry })
+    } else if (geometry instanceof PolygonGeometry) {
+      this.renderPolygon({ appearance, geometry })
+    } else if (geometry instanceof TextGeometry) {
+      this.renderText({ appearance, camera, geometry })
     }
```
MonicaOlejniczak: Yeah probably should've added that haha

src/client/canvas/object/scene.ts
```diff
@@ -0,0 +1,42 @@
...
+import { Transform } from '../../math/transform'
+import { Group } from './group'
+import { GroupOpts } from './group'
+import { Object2d } from './object2d'
+import { action } from 'mobx'
+import { observable } from 'mobx'
+
+export type SceneOpts = GroupOpts
+
+export class Scene implements Object2d {
```
MonicaOlejniczak: I was thinking about this myself. The only things threejs adds to their scene object that isn't on an object3d is an overall background, fog and an override. I doubt we'll need any of those and they could just be applied to a parent shape appearance in our case. I do like referring to the root object as a scene type but that's literally it's only benefit. Thoughts on the latter?

src/client/canvas/renderer.ts
MonicaOlejniczak: Yeah, I just got a tad lazy

src/client/canvas/renderer.ts
```diff
@@ -43,19 +43,36 @@ export class CanvasRenderer {
...
+        translate: {
+          x: 0,
+          y: 0,
+        },
+      })
+      this.context.scale(obj.transform.scale.x, obj.transform.scale.y)
+      this.context.rotate(-obj.transform.rotate)
+      this.context.translate(translateDash.x, translateDash.y)
+      if (obj instanceof Group) {
+        this.renderObjects(obj.children, camera)
```
BrendanAnnable: I think this is simply a case of:
`this.renderShape(obj, camera.clone().applyTransform(obj.transform))`

src/client/canvas/object/scene.ts
BrendanAnnable: I think `const scene = Group.of()` is sufficient for that purpose 🙂

PR #100, opened 2017-07-23
Add Dashboard

MonicaOlejniczak:
image

- An aggregate of robots on the field is shown in the top half of the screen on the canvas.
  - The line from the robot to the ball indicates where the robot thinks the ball is.
  - Arrows represent the kick target location from the ball.
- Robot details are shown in the bottom half of the screen.
  - Displays game mode, game phase, behaviour, penalty and icons representing the time the last camera image, last seen ball and goals were received.
  - A warning icon appears next to the penalty for penalised robots.
  - Icons flash red when an item has not been received for a long time.

PR #107, opened 2017-07-24
[WIP] Add localisation objects

JosephusPaye:
- [ ] Add ball
- [ ] Add ball confidence disk
- [ ] Add robot confidence disk


src/client/components/localisation/view_model.ts
```diff
@@ -52,6 +55,13 @@ export class LocalisationViewModel {
   }
 
   @computed
+  private get balls(): Object3D[] {
+    return this.model.robots
+      .filter(robotModel => robotModel.visible)
```
BrendanAnnable: 🚗 🔫  If we add a new `@computed` method:

```ts
@computed
private get visibleRobots(): Object3D[] {
     return this.model.robots.filter(robotModel => robotModel.visible)
}
```

Then this can be:
```ts
@computed
private get balls(): Object3D[] {
  this.visibleRobots.map(robotModel => RobotViewModel.of(robotModel).ball)
}
```

and `robots()` could use it too.

src/client/components/localisation/darwin_robot/view_model.ts
```diff
@@ -26,4 +27,17 @@ export class RobotViewModel {
...
     return robot
   }
+
+  @computed
+  public get ball(): Object3D {
+    const ball = new Object3D()
+
+    ball.position.x = this.model.ball.position.x
+    ball.position.y = this.model.ball.position.y
+    ball.position.z = this.model.ball.position.z
```
BrendanAnnable: 🚗 🔫  I think you could move these positions directly into `BallViewModel` and just have `return BallViewModel.of(this.model.ball).ball` here

src/client/components/localisation/ball/model.ts
```diff
@@ -0,0 +1,32 @@
...
+    this.radius = radius
+    this.color = color
+  }
+
+  public static of(opts: BallModelOpts = {}) {
+    return new BallModel(
+      opts.name || 'Ball',
+      opts.position || Vector3.of(),
+      opts.radius || 0.0335,
+      opts.color || '#FFCC00',
```
BrendanAnnable: 🚗 🔫  We recently switched to using [this syntax](https://github.com/NUbots/NUsight2/pull/100/commits/3471a7440804f21934ea294bd16e0e13e444eabe) for default values.

Allows us to be consistent for numerical and "falsey" values, e.g. `scale || 1` will actually return `1` if scale has a value of `0`, where as using `{ scale = 1 }` will only default it if `scale === undefined`.

src/client/components/localisation/view_model.ts
Phasedd: yeah I had a similar thought, but instead I was going to make `get robot(): Object3D[]` to return `LocalisationRobotModel[]` and do this:

```TS
this.robots.forEach(ball => {
  scene.add(robot.robot)
  scene.add(robot.ball)
})
```

TrentHouliston: I think the grouping should be LocalisationRobotModels, since the ball and robot are tightly coupled, wouldn't ever make sense to show balls without the attached robots.

src/client/components/localisation/darwin_robot/view_model.ts
JosephusPaye: Moved!

src/client/components/localisation/ball/model.ts
JosephusPaye: Changed!

src/client/components/localisation/ball/view_model.ts
```diff
@@ -0,0 +1,42 @@
...
+    return new Mesh(geometry, material)
+  }
+
+  @computed
+  public get ball(): Object3D {
+    const ball = new Object3D()
+
+    ball.position.x = this.model.position.x
+    ball.position.y = this.model.position.y
+    ball.position.z = this.model.position.z
```
BrendanAnnable: Oh I probably didn't explain very well, you can move these `.position` properties directly onto the mesh itself, example: https://github.com/NUbots/NUsight2/blob/master/src/client/components/localisation/darwin_robot/body/view_model.ts#L23

src/client/components/localisation/confidence_ellipse/model.ts
```diff
@@ -0,0 +1,46 @@
...
+  @observable public rotationAngle: number
+
+  public constructor(opts: Partial<ConfidenceEllipseModel>) {
+    Object.assign(this, opts)
+  }
+
+  public static of({
+    color = '#0000ff',
+    opacity = 0.5,
+    position = new Vector3(0, 0, 0.001123),
```
BrendanAnnable: 🚗 🔫  Can probably move the offset concern to the view model by using `polygonOffset` on the material.

Example: http://jsfiddle.net/adt1d5b7/1/

Reference: https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glPolygonOffset.xhtml

Disclaimer: I haven't used it myself, but I have been meaning to solve this issue, and will probably use it for the field lines.

BrendanAnnable:
Is this still a thing?

JosephusPaye:
Ping @TrentHouliston

TrentHouliston:
Yes this should still be a thing

JosephusPaye:
This is stale now, would be better to do it from scratch if we get around to it.

PR #108, opened 2017-07-25
Add anticlockwise property to Transform

BrendanAnnable:
Because we flip the camera along the Y axis to bring the NUbots coordinate system into the native canvas coordinate system, we need the ability to also specify the right-handedness of the camera transform to ensure that rotations are still applied counter-clockwise.


BrendanAnnable:
Hmm this is still not quite correct. I'll look at this again later today. Vector2 needs updating as well.

BrendanAnnable:
It is finally working!

PR #109, opened 2017-07-25
Update NUClearNet.js

Phasedd:
Update NUClearNet.js to 1.5.2


BrendanAnnable:
I don't know much about the new NUClearNet version. Where did the value `10.1.255.255` come from? Why is that needed now?

Phasedd:
`10.1.255.255` is the broadcast address.

For now it may be a robocup specific implementation if we find a better way so we can toggle between broadcast and multicast.

If we don't want to merge this change in, we can just keep it updated to master so we can use it at robocup.

PR #112, opened 2017-07-25
Update field dimensions for RoboCup 2017

JosephusPaye:
Based on new data here: https://github.com/NUbots/NUbots/pull/92, which includes physical measurements of the field.


Phasedd:
Me and @JosephusPaye used a tape measure on the field.

PR #114, opened 2017-07-26
Add Matrix2 and Matrix3 models

TrentHouliston:
These are needed to parse some of the messages that will be coming in the overview packet

### DIFF
--- a/src/client/math/matrix2.ts
+++ b/src/client/math/matrix2.ts
@@ -21,7 +21,7 @@ export class Matrix2 {
     y?: { x?: number, y?: number }
   } | null): Matrix2 {
     if (!mat) {
-      mat = { x: { x:0, y:0, z:0 }, y: { x:0, y:0, z:0 }, z: { x:0, y:0, z:0 } }
+      mat = { x: { x:0, y:0 }, y: { x:0, y:0 } }
     }
     return new Matrix2(Vector2.from(mat.x), Vector2.from(mat.y))
   }
--- a/src/client/math/matrix2.ts
+++ b/src/client/math/matrix2.ts
@@ -1,0 +1,79 @@
+import { action } from 'mobx'
+import { computed } from 'mobx'
+import { observable } from 'mobx'
+import { Vector2 } from './vector2'
+
+export class Matrix2 {
+  @observable public x: Vector2
+  @observable public y: Vector2
+
+  public constructor(x: Vector2, y: Vector2) {
+    this.x = x
+    this.y = y
+  }
+
+  public static of() {
+    return new Matrix2(new Vector2(0, 0), new Vector2(0, 0))
+  }
+
+  public static from(mat?: {
+      x?: { x?: number, y?: number },
+      y?: { x?: number, y?: number },
+    } | null): Matrix2 {
+    if (!mat) {
+      mat = { x: { x:0, y:0, z:0 }, y: { x:0, y:0, z:0 }, z: { x:0, y:0, z:0 } }
+    }
+    return new Matrix2(Vector2.from(mat.x), Vector2.from(mat.y))
+  }
+
+  @action
+  public set(x: Vector2, y: Vector2): Matrix2 {
+    this.x = x
+    this.y = y
+    return this
+  }
+
+  @action
+  public clone(): Matrix2 {
+    return new Matrix2(this.x.clone(), this.y.clone())
+  }
+
+  @action
+  public copy(v: Matrix2): Matrix2 {
+    this.x = v.x
+    this.y = v.y
+    return this
+  }
+
+  @action
+  public multiplyScalar(scalar: number): Matrix2 {
+    this.x.multiplyScalar(scalar)
+    this.y.multiplyScalar(scalar)
+    return this
+  }
+
+  @action
+  public divideScalar(scalar: number): Matrix2 {
+    this.x.divideScalar(scalar)
+    this.y.divideScalar(scalar)
+    return this
+  }
+
+  @action
+  public add(movement: Matrix2): Matrix2 {
+    this.x.add(movement.x)
+    this.y.add(movement.y)
+    return this
+  }
+
+  @action
+  public subtract(movement: Matrix2): Matrix2 {
+    this.x.subtract(movement.x)
+    this.y.subtract(movement.y)
+    return this
+  }
+
+  @computed get trace(): number {
+    return this.x.x + this.y.y
+  }
+}
--- a/src/client/math/matrix2.ts
+++ b/src/client/math/matrix2.ts
@@ -44,8 +44,8 @@ export class Matrix2 {
 
   @action
   public copy(m: Matrix2): Matrix2 {
-    this.x = m.x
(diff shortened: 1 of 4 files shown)


src/client/math/matrix2.ts
```diff
@@ -0,0 +1,79 @@
...
+    this.y.add(movement.y)
+    return this
+  }
+
+  @action
+  public subtract(movement: Matrix2): Matrix2 {
+    this.x.subtract(movement.x)
+    this.y.subtract(movement.y)
+    return this
+  }
```
BrendanAnnable: What do you think of letting this class be agnostic of its application and removing any methods that aren't applicable to generic matrices? I think we can remove `add`/`subtract` on `Matrix3` / `Matrix4` in favour of a generic `.multiply` 🙂

src/client/math/matrix2.ts
```diff
@@ -0,0 +1,79 @@
...
+  @observable public x: Vector2
+  @observable public y: Vector2
+
+  public constructor(x: Vector2, y: Vector2) {
+    this.x = x
+    this.y = y
+  }
+
+  public static of() {
+    return new Matrix2(new Vector2(0, 0), new Vector2(0, 0))
```
BrendanAnnable: Probably should default to an identify matrix here.

src/client/math/matrix3.ts
```diff
@@ -0,0 +1,88 @@
...
+  @observable public z: Vector3
+
+  public constructor(x: Vector3, y: Vector3, z: Vector3) {
+    this.x = x
+    this.y = y
+    this.z = z
+  }
+
+  public static of() {
+    return new Matrix3(new Vector3(0, 0, 0), new Vector3(0, 0, 0), new Vector3(0, 0, 0))
```
BrendanAnnable: Probably should default to an identify matrix here.

src/client/math/matrix3.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  @action
+  public subtract(movement: Matrix3): Matrix3 {
+    this.x.subtract(movement.x)
+    this.y.subtract(movement.y)
+    this.z.subtract(movement.z)
+    return this
+  }
+
+  @computed get trace(): number {
```
BrendanAnnable: I think it would make sense to keep `@computed` properties above `@action` methods 🙂

src/client/math/matrix2.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  @action
+  public subtract(movement: Matrix2): Matrix2 {
+    this.x.subtract(movement.x)
+    this.y.subtract(movement.y)
+    return this
+  }
+
+  @computed get trace(): number {
```
BrendanAnnable: I think it would make sense to keep `@computed` properties above `@action` methods 🙂

src/client/math/matrix2.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  public static of() {
+    return new Matrix2(new Vector2(0, 0), new Vector2(0, 0))
+  }
+
+  public static from(mat?: {
+    x?: { x?: number, y?: number },
+    y?: { x?: number, y?: number }
+  } | null): Matrix2 {
+    if (!mat) {
```
BrendanAnnable: I'm pretty hesitant of this functionality here (I realise its for consistency with Vector2). I'm not sure which situations we would expect to provide a `null` object, but they _should_ be rare if at all. I think I'd prefer a type error instead. i.e. I think we should just make everything required instead by removing the `?` and this conditional.

src/client/math/matrix2.ts
TrentHouliston: We can add it later if we need it, however i'm somewhat of the opinion that these classes are models and should have limited operations on them. If you needed to do more complex operations like matrix multiplication you'd be better off doing it in a system like threejs.

src/client/math/matrix2.ts
TrentHouliston: I did a sorting, but picked the wrong one!

src/client/math/matrix2.ts
TrentHouliston: Should the vector type also not have this?

src/client/math/matrix3.ts
```diff
@@ -0,0 +1,88 @@
...
+    return new Matrix3(new Vector3(1, 0, 0), new Vector3(0, 1, 0), new Vector3(0, 0, 1))
+  }
+
+  public static from(mat?: {
+    x?: { x?: number, y?: number, z? :number },
+    y?: { x?: number, y?: number, z? :number },
+    z?: { x?: number, y?: number, z? :number }
+  } | null): Matrix3 {
+    if (!mat) {
+      mat = { x: { x:0, y:0, z:0 }, y: { x:0, y:0, z:0 }, z: { x:0, y:0, z:0 } }
```
BrendanAnnable: `return Matrix3.of()`

src/client/math/vector3.ts
```diff
@@ -17,10 +18,21 @@ export class Vector3 {
     return new Vector3(0, 0, 0)
   }
 
+  public static from(vec3?: { x?: number, y?: number, z?: number } | null): Vector3 {
+    if (!vec3) {
+      vec3 = { x: 0, y: 0, z: 0 }
```
BrendanAnnable: `return Vector3.of()`

src/client/math/vector3.ts
```diff
@@ -77,5 +89,13 @@ export class Vector3 {
     this.z += movement.z
     return this
   }
+
+  @action
+  public subtract(movement: Vector3): Vector3 {
```
BrendanAnnable: `s/movement/v`

src/client/math/vector3.ts
```diff
@@ -1,6 +1,7 @@
 import { action } from 'mobx'
 import { computed } from 'mobx'
 import { observable } from 'mobx'
+import { Vector2 } from './vector2'
```
BrendanAnnable: unused

src/client/math/matrix2.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  @action
+  public clone(): Matrix2 {
+    return new Matrix2(this.x.clone(), this.y.clone())
+  }
+
+  @action
+  public copy(m: Matrix2): Matrix2 {
+    this.x = m.x
+    this.y = m.y
```
BrendanAnnable: this currently copied the reference, this needs to be:

```
this.x.copy(m.x)
this.y.copy(m.y)
```

src/client/math/matrix3.ts
```diff
@@ -0,0 +1,56 @@
...
+  @action
+  public clone(): Matrix3 {
+    return new Matrix3(this.x.clone(), this.y.clone(), this.z.clone())
+  }
+
+  @action
+  public copy(m: Matrix3): Matrix3 {
+    this.x = m.x
+    this.y = m.y
+    this.z = m.z
```
BrendanAnnable: these currently copies the reference as well

PR #116, opened 2017-07-26
Make travis much faster by caching node_modules 🏎🏎🏎

TrentHouliston:
But if the yarn.lock file changes we purge the cache to make sure we don't get unexpected build errors.


TrentHouliston:
It would work, but it seems that yarn deletes the yarn.lock file I put into the node_modules folder. It will need to be stored elsewhere.

TrentHouliston:
Tested! It deletes the cache both ways!

PR #119, opened 2017-07-27
[WIP] Add vision

BrendanAnnable:
Proof-of-concept using a react/declarative approach 😄 Renders the green box using a WebGL canvas and the orange circle using the Canvas 2D API. Currently pretty rough around the edges since I just hacked on it, but should give you an idea of the approach I had in my mind 🙂

screen shot 2017-07-27 at 8 52 17 pm


LukeFarrawell:
Just a heads up, there needs to be support for both pinhole and radial camera support.

On that note too, the same data comes for both the pinhole and radial camera, take a look at at [VisionController.js](https://github.com/NUbots/NUsight/blob/develop/public/app/view/window/VisionController.js#L705) in NUsight on how the calculations are performed.

The same for [balls](https://github.com/NUbots/NUsight/blob/develop/public/app/view/window/VisionController.js#L784) and also for the the [horizon](https://github.com/NUbots/NUsight/blob/develop/public/app/view/window/VisionController.js#L568)

BrendanAnnable:
Ah right. Would it be possible to get recordings for each type of robot + camera combination? I can't really test against anything without some real data.

LukeFarrawell:
I won't be able to for a week

TrentHouliston:
Closed in favour of #176 and #182

PR #122, opened 2017-07-28
[Dashboard] Indicate robot connection status

BrendanAnnable:
This PR has two changes:

- Add connection status icons to each robot panel in the dashboard:
screen shot 2017-07-28 at 4 26 28 pm
screen shot 2017-07-28 at 4 26 08 pm

- Hides robots from the field if they are not connected, even if they are enabled.

Since the current virtual robots are always connected (something we might want change in the future), the easiest way to test this is with real networking, i.e. `yarn start` and `yarn simulate`.


src/client/base/browser_clock.ts
```diff
@@ -0,0 +1,31 @@
...
+  const handle = window.setTimeout(cb, 0)
+  return window.clearTimeout.bind(null, handle)
+}
+
+function performanceNow() {
+  return window.performance.now() * 1e-3
+}
+
+
+export const BrowserSystemClock: Clock = {
```
BrendanAnnable: Hm I was planning to use this for fading them out over time, but I excluded that from this PR. So this had nothing to do with the added changes but we will surely be needing this one anyway.

src/client/components/dashboard/dashboard_robot/model.ts
```diff
@@ -108,12 +108,17 @@ export class DashboardRobotModel {
...
+    return this.robot.connected
+  }
+
+  @computed
   public get name(): string {
     return this.robot.name
   }
 
   @computed
-  public get visible(): boolean {
```
BrendanAnnable: Moved the "visibility" to be a view model concern, since it now depends where it is shown.

src/client/components/app/network.ts
```diff
@@ -38,6 +38,7 @@ export class AppNetwork {
     }
   }
 
+  @action
```
BrendanAnnable: I'm surprised anything has been working without this 🙂 Maybe it will fix the disappearing field issue ¯\_(ツ)_/¯

JosephusPaye: Hmmm

PR #123, opened 2017-07-29
Add BrowserSystemClock

BrendanAnnable:
Adds `BrowserSystemClock`, almost identical to `NodeSystemClock`, to be used in a browser context.

Also moves the common `Clock` interface and `FakeClock` into `shared/time`.


src/client/time/browser_clock.ts
```diff
@@ -0,0 +1,29 @@
+import { CancelTimer } from '../../shared/time/clock'
+import { Clock } from '../../shared/time/clock'
+
+function setTimeout(cb: (...args: any[]) => void, seconds: number): CancelTimer {
+  const handle = window.setTimeout(cb, seconds * 1e3)
+  return window.clearTimeout.bind(null, handle)
```
MonicaOlejniczak: Can you change this to undefined? That is the preferred in typescript and is consistent with the rest of the codebase. The same for the one's below.

src/client/time/browser_clock.ts
```diff
@@ -0,0 +1,29 @@
+import { CancelTimer } from '../../shared/time/clock'
+import { Clock } from '../../shared/time/clock'
+
+function setTimeout(cb: (...args: any[]) => void, seconds: number): CancelTimer {
+  const handle = window.setTimeout(cb, seconds * 1e3)
```
MonicaOlejniczak: Not a huge fan of the repeated constant. Either make it a variable or implement conversion functions.

BrendanAnnable: Done!

src/client/time/browser_clock.ts
BrendanAnnable: Done!

PR #132, opened 2017-10-27
Upgrade to React 16

BrendanAnnable:
Upgraded ~a few other things~ everything else while I was there.


webpack.config.ts
```diff
@@ -49,7 +49,7 @@ export default {
         use: isProduction
             ? 'awesome-typescript-loader?module=es6'
             : [
-              'react-hot-loader',
+              'react-hot-loader/webpack',
```
BrendanAnnable: https://github.com/gaearon/react-hot-loader#typescript

PR #127, opened 2017-07-30
[WIP] Chart

MonicaOlejniczak:
:)

PR #133, opened 2017-10-27
Make the camera act like a camera

BrendanAnnable:
Moving a camera to the left should move the objects it is looking at to the right.

Taking this out of https://github.com/NUbots/NUsight2/pull/127 so I can use it for other things!

### DIFF
--- a/src/client/components/dashboard/field/controller.ts
+++ b/src/client/components/dashboard/field/controller.ts
@@ -10,19 +10,19 @@ export class FieldController {
   public onFieldResize(model: FieldModel, width: number, height: number) {
     const fieldWidth = model.fieldWidth
     const fieldLength = model.fieldLength
-    const scaleX = width / fieldLength
-    const scaleY = height / fieldWidth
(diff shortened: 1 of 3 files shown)


src/client/components/dashboard/field/controller.ts
```diff
@@ -10,19 +10,19 @@ export class FieldController {
...
+    const scaleX = fieldLength / width
+    const scaleY = fieldWidth / height
 
-    const canvasAspect = height / width
-    const fieldAspect = fieldWidth / fieldLength
+    const canvasAspect = width / height
+    const fieldAspect = fieldLength / fieldWidth
     // Ensures the canvas is scaled such that the entire field is always visible, regardless of aspect ratio.
-    const scale = canvasAspect < fieldAspect ? scaleY : scaleX
+    const scale = canvasAspect > fieldAspect ? scaleY : scaleX
```
BrendanAnnable: Scale is the same (the inverse).

PR #141, opened 2017-11-26
Replace 'of' with 'create'

BrendanAnnable:
I was asked why `of`? and I didn't really have an answer. I think `create` does make it a bit more intuitive.

Potentially more distracting though.


BrendanAnnable:
Looking at some of the diffs, its actually much more distracting imo. Not worth it.

Closing, with the expectation that we'll have to explain that `of` means `create` to anyone starting to work on the codebase.

JosephusPaye:
I think `of` makes sense, since we're passing in arguments and want to create something _**of**_ those arguments. 🤔

BrendanAnnable:
Excellent! 😄

PR #142, opened 2017-11-29
Uninstall unimplemented features

BrendanAnnable:
I think it might be good to uninstall these for now, and then get excited as we see them pop back as they get implemented 🙂


TrentHouliston:
Not going to also delete the modules themselves? there's not much there anyway

BrendanAnnable:
They have icons and things!

BrendanAnnable:
We should probably delete anything we have no plan to implement in NUsight2, do any of these fall under that category?

BrendanAnnable:
Future component writers, steal the icons from this PR!

PR #144, opened 2017-11-30
Improve streaming performance of WebSocketProxy

TrentHouliston:


src/server/nuclearnet/web_socket_proxy_nuclearnet_server.ts
```diff
@@ -163,20 +162,29 @@ class PacketProcessor {
...
-    return () => {
+    return (processing: number) => {
       if (!isDone) {
-        const eventQueueSize = this.eventQueueSize.get(event) || 0
-        this.eventQueueSize.set(event, eventQueueSize - 1)
+        // Calculate network latency
+        const latency = this.clock.performanceNow() - start
+
+        // Update our performance tracking information
+        const tracking = this.eventTracking.get(event) || { active: 0, latency: 1.0, processing: 0.5 }
```
TrentHouliston: This default shouldn't be here as it should be impossible to get here without having sent a packet

BrendanAnnable: You can use `this.eventTracking.get(event)!` in that case

ysims:
Old and unlikely to be finished. Can be recoded if needed in future.

PR #148, opened 2018-01-11
Fix the hash function when the result isn't 16 characters

TrentHouliston:
When the hash code isn't 16 characters (because the MSBs are to small) the hash code was too short and broke everything.
This zero pads the resulting string to ensure it's always 16 characters.

### DIFF
--- a/src/server/nuclearnet/fake_nuclearnet_server.ts
+++ b/src/server/nuclearnet/fake_nuclearnet_server.ts
@@ -86,7 +86,7 @@ export function hashType(type: string): Buffer {
   // See https://goo.gl/6NDPo2
   let hashString: string = XXH.h64(type, 0x4e55436c).toString(16)
   // The hash string may truncate if it's smaller than 16 characters so we pad it with 0s
-  hashString = `0000000000000000${hashString}`.slice(-16)
+  hashString = ('0'.repeat(16) + hashString).slice(-16)
 
   return Buffer.from((hashString.match(/../g) as string[]).reverse().join(''), 'hex')
 }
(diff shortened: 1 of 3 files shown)


src/server/nuclearnet/fake_nuclearnet_server.ts
```diff
@@ -84,6 +84,9 @@ export class FakeNUClearNetServer {
 export function hashType(type: string): Buffer {
   // Matches hashing implementation from NUClearNet
   // See https://goo.gl/6NDPo2
-  const hashString: string = XXH.h64(type, 0x4e55436c).toString(16)
+  let hashString: string = XXH.h64(type, 0x4e55436c).toString(16)
+  // The hash string may truncate if it's smaller than 16 characters so we pad it with 0s
+  hashString = `0000000000000000${hashString}`.slice(-16)
```
BrendanAnnable: How about `('0'.repeat(16) + hashString).slice(-16)`, so you can see both need to be 16?

BrendanAnnable:
I think these tests would be sufficient for now:

```ts
import { hashType } from '../fake_nuclearnet_server'

describe('FakeNUClearNetServer', () => {
  describe('#hashType', () => {
    it('matches previous snapshots', () => {
      expect(hashType('foo')).toMatchSnapshot()
      expect(hashType('bar')).toMatchSnapshot()
      expect(hashType('baz')).toMatchSnapshot()
      expect(hashType('cheesecake')).toMatchSnapshot()
    })

    it('ensures it returns 8 bytes', () => {
      expect(hashType('foo').byteLength).toBe(8)
      expect(hashType('bar).byteLength).toBe(8)
      expect(hashType('baz').byteLength).toBe(8)
      // cheesecake happens to hash to 7 bytes without padding.
      expect(hashType('cheesecake').byteLength).toBe(8)
    })
  })
})
```

PR #149, opened 2018-01-12
Add many tslint rules that generally seem useful

TrentHouliston:
Also extend from tslint:latest so any new recommended rules that come out will be applied (and if they break stuff we can manipulate them then).

https://palantir.github.io/tslint/rules/


src/server/nbs/nbs_frame_chunker.ts
```diff
@@ -32,7 +33,9 @@ export class NbsFrameChunker extends stream.Transform {
     this.buffers.push(chunk)
 
     let frame
+    /* tslint:disable no-conditional-assignment */
```
BrendanAnnable: `/* tslint:disable-next-line no-conditional-assignment */` should work here

PR #150, opened 2018-01-12
Add the goal and penalty mark field lines

LukeFarrawell:
image

### DIFF
--- a/src/client/components/localisation/field/view_model.ts
+++ b/src/client/components/localisation/field/view_model.ts
@@ -63,6 +63,7 @@ export class FieldViewModel {
     const goalAreaLength = this.model.dimensions.goalAreaLength
     const goalDepth = this.model.dimensions.goalDepth
     const goalWidth = this.model.dimensions.goalWidth
+    const penaltyMarkDistance = this.model.dimensions.penaltyMarkDistance
 
     const halfLength = fieldLength * 0.5
     const halfWidth = fieldWidth * 0.5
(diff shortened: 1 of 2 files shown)


src/client/components/localisation/field/view_model.ts
```diff
@@ -89,8 +119,9 @@ export class FieldViewModel {
   }
 
   private buildRectangle(x: number, y: number, w: number, h: number, lw: number) {
-    const x1 = x - lw * 0.5
-    const x2 = x + w + lw * 0.5
+    let x1 = x - lw * 0.5
+    let x2 = x + w + lw * 0.5
```
BrendanAnnable: Any reason for making these use `let` instead of `const`?

yarn.lock
```diff
@@ -2902,14 +2902,10 @@ extract-text-webpack-plugin@^3.0.0:
     schema-utils "^0.3.0"
     webpack-sources "^1.0.1"
 
-extsprintf@1.3.0:
+extsprintf@1.3.0, extsprintf@^1.2.0:
```
BrendanAnnable: Are these `yarn.lock` changes necessary? Can we just remove them from this PR if not? I'm not sure what caused this, but make sure `yarn --version` is at least >= 1.x.x.

LukeFarrawell: I had a yarn version < 1.x.x, it seems that was the issue.

PR #152, opened 2018-01-13
SVG Renderer

TrentHouliston:
A renderer for the canvas drawing system that uses SVGs instead of cavas.


src/client/canvas/svg/svg.tsx
```diff
@@ -0,0 +1,82 @@
...
+      strokeDashoffset: appearance.lineDashOffset,
+      strokeWidth: appearance.lineWidth,
+      stroke: appearance.strokeStyle,
+    }
+  } else {
+    throw new Error(`Unsupported appearance type ${appearance}`)
+  }
+}
+
+export function transform(transform: Transform): string {
```
BrendanAnnable: `transform` is a very common variable name and will conflict with the rest of my suggestions, can these two above functions be named `svgTransform` and `svgApperance`?

src/client/canvas/svg/text.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+import { appearance, transform } from './svg'
+
+@observer
+export class Text extends Component<{model: Shape<TextGeometry>, world: Transform}> {
+
+  public render() {
+    // TODO handle font size and max width properly
+
+    const m = this.props.model
```
BrendanAnnable: Can we just restructuring instead of 1 letter variable names? Here is how I would write this class:

```ts
export class Text extends Component<{ model: Shape<TextGeometry>, world: Transform }> {
  public render() {
    const { world, model: { geometry, appearance } } = this.props
    const transform = Transform.of(geometry.alignToView ? {
      scale: { x: Math.sign(world.scale.x), y: Math.sign(world.scale.y) },
      rotate: world.rotate,
    } : undefined)

    return (
      <text
        x={geometry.x}
        y={geometry.y}
        fontFamily={geometry.fontFamily}
        fontSize={'0.015rem'} // TODO handle font size and max width properly
        textAnchor={geometry.textAlign}
        dominantBaseline={geometry.textBaseline}
        transform={toSvgTransform(transform)}
        {...toSvgAppearance(appearance)}
      >
        {geometry.text}
      </text>
    )
  }
}
```

If you agree, please make sure all the others are also consistent 🙂

src/client/canvas/svg/group.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { transform, viewForGeometry } from './svg'
+
+@observer
+export class Group extends Component<{model: GroupGeometry, world: Transform}> {
+
+  public render() {
+    return <g transform={transform(this.props.model.transform)}>
+      {this.props.model.children.map((obj, i) =>
+        viewForGeometry(obj, i, this.props.world.clone().then(this.props.model.transform.inverse())))}
+    </g>
```
BrendanAnnable: ⛳️ ing:

```ts
type Props = { model: GroupGeometry, world: Transform }
export const Group = observer(({ model: { children, transform }, world }: Props) => (
  <g transform={toSvgTransform(transform)}>
    {children.map((child, i) => viewForGeometry(child, i, world.clone().then(transform)))}
  </g>
))
```

We can apply this to every react component, since they are all stateless.

src/client/canvas/svg/svg.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const r = (180.0 / Math.PI) * transform.rotate // SVG rotations are in degrees
+  const t = transform.translate
+  return `translate(${t.x}, ${t.y}) rotate(${r}) scale(${s.x}, ${s.y})`
+}
+
+export function viewForGeometry(obj: Object2d, index: number, world: Transform): JSX.Element {
+
+  // TODO wrap things in appearance tags
+  if (obj instanceof GroupGeometry) {
+    return <Group key={index} model={obj} world={world} />
```
BrendanAnnable: Having the `key` prop set here forces that these components can only be rendered within a list.

Moving the responsibility for setting the `key` prop to whoever is rendering the list would be ideal here.

In order to achieve that, the return type of this function must be a renderable `ComponentType`, not a `JSX.Element`.

I'll have a spike and post the gist back here.

src/client/components/dashboard/field/view.tsx
```diff
@@ -1,69 +1,52 @@
...
-    }
-    const context: CanvasRenderingContext2D = this.field.getContext('2d')!
-    this.renderer = CanvasRenderer.of(context)
-    this.stopAutorun = autorun(() => this.renderField())
-    this.rafId = requestAnimationFrame(this.onAnimationFrame)
+  public render() {
+    const model = this.props.model
+    const viewModel = FieldViewModel.of(model)
+    return <div className={style.container}>
+      <ReactResizeDetector handleWidth handleHeight onResize={this.onResize}/>
```
BrendanAnnable: Woohoo! 🎉

src/client/canvas/object/shape.ts
```diff
@@ -24,24 +24,24 @@ export type Geometry =
...
 
-export type ShapeOpts = {
+export type ShapeOpts<T extends Geometry> = {
   appearance: Appearance
-  geometry: Geometry
+  geometry: T
 } & GroupOpts
 
-export class Shape implements Object2d {
+export class Shape<T extends Geometry> implements Object2d {
```
BrendanAnnable: What was the reason for making this generic?

TrentHouliston: When I use them later on as models in the SVG renders they go on the props. Having the type available there so I don't have to make new types is useful.

src/client/canvas/svg/svg.tsx
TrentHouliston: 👍

src/client/canvas/svg/svg.tsx
BrendanAnnable: Have a look at [this gist](https://gist.github.com/BrendanAnnable/0d8e2539dbdcc617dbd4f12aeeae2c1d).

src/client/canvas/svg/arc.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  private arc(origin: {x: number, y: number},
+              r: number,
+              a0: number,
+              a1: number,
+              ccw: boolean): string {
+
+    // Is this arc empty?
+    if (r === 0) {
+      return ''
```
BrendanAnnable: Should probably move this into `render()` so that a path isn't even rendered.

e.g. `render() { return r !== 0 && <path d={...}/> }`

src/client/canvas/svg/arc.tsx
```diff
@@ -0,0 +1,67 @@
...
+    // Work out our start and end points on the circle
+    const p0 = {
+      x: origin.x + r * Math.cos(a0),
+      y: origin.y + r * Math.sin(a0),
+    }
+
+    const p1 = {
+      x: origin.x + r * Math.cos(a1),
+      y: origin.y + r * Math.sin(a1),
+    }
```
BrendanAnnable: Can we use `Vector2` for this maths? And add a `Vector2.fromPolar(radius, angle)` static method to help this read nicely.

src/client/canvas/object/shape.ts
```diff
@@ -24,24 +24,24 @@ export type Geometry =
...
-  constructor(opts: ShapeOpts) {
+  constructor(opts: ShapeOpts<T>) {
     this.appearance = opts.appearance
     this.geometry = opts.geometry
     this.group = Group.of(opts)
   }
 
   public static of(geometry: Geometry, appearance: Appearance = BasicAppearance.of()) {
-    return new Shape({
+    return new Shape<Geometry>({
```
BrendanAnnable: You should template this method:

```ts
  public static of<T extends Geometry>(geometry: T, appearance: Appearance = BasicAppearance.of()) {
    return new Shape<T>({
      ...
    })
  }
```

src/client/canvas/svg/arc.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+    // Our cw value must be a 0/1 not true/false
+    const cw = +!ccw
+    const da = ccw ? a0 - a1 : a1 - a0
+
+    // Move to (x0,y0).
+    let output = `M${p0.x} ${p0.y}`
+
+    // Draw the arc
+    output += `A${r} ${r} 0 ${+(da >= Math.PI)} ${cw} ${p1.x} ${p1.y}`
```
BrendanAnnable: Some of these render functions are pretty complicated, can we add some snapshot tests? They are pretty simple because you don't have to write the expected value manually. Here is one for circle:

```ts
import * as React from 'react'
import { BasicAppearance } from '../../appearance/basic_appearance'
import { CircleGeometry } from '../../geometry/circle_geometry'
import { Shape } from '../../object/shape'
import { Circle } from '../circle'

describe('CircleGeometry', () => {
  it('renders', () => {
    const model = Shape.of(CircleGeometry.of({
      radius: 2,
    }), BasicAppearance.of())
    expect(<Circle model={model}/>).toMatchSnapshot()
  })
})
```

TrentHouliston: I'm not sure snapshots are the best way to make tests for these. It is valid to draw these in a different way which won't match the snapshot. E.g. if I changed the way I drew arcs to have a clippled circle for example, it wouldn't match the snapshot but it would be the same visually.

BrendanAnnable: The snapshots test serve to ensure that any future changes to the output are reviewed and intentional. If you wanted to change the way your drew arcs, you would just update the snapshots. Its mainly about allowing non-output changing refactors such that you are confident that the rendered output did not change from whatever it was before.

The type of visual regression testing you're proposing would be something achievable with something like [Percy](https://percy.io/), but in the mean time I think this is the next best thing.

src/client/canvas/svg/text.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+  return <text
+    x={x}
+    y={y}
+    fontFamily={fontFamily}
+    fontSize={'0.015rem'}
+    textAnchor={textAlign}
+    dominantBaseline={textBaseline}
+    transform={toSvgTransform(t)}
+    {...svgAppearanceAttributes(appearance)}>
```
BrendanAnnable: Can we move the trailing `>` to the next line, such that appending a new prop to this list wouldn't result in a git diff of two lines instead of one?

src/client/canvas/renderer.ts
```diff
@@ -13,7 +13,7 @@ import { PolygonGeometry } from './geometry/polygon_geometry'
 import { TextGeometry } from './geometry/text_geometry'
 import { Group } from './object/group'
 import { Object2d } from './object/object2d'
-import { Shape } from './object/shape'
+import { Geometry , Shape } from './object/shape'
 
 export class CanvasRenderer {
```
BrendanAnnable: Can we rename this file to `canvas_renderer`?

src/client/canvas/svg_renderer.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { toSvgTransform } from './svg/svg'
+
+export type SVGRendererProps = {
+  className: string
+  scene: GroupGeometry
+  camera: Transform
+}
+
+@observer
+export class SVGRenderer extends Component<SVGRendererProps> {
```
BrendanAnnable:
```ts
export const SVGRenderer = observer(({ className, scene, camera }: SVGRendererProps) => (
  <svg className={className}>
    <g transform={toSvgTransform(camera.inverse())}>
      <Group model={scene} world={camera.inverse()}/>
    </g>
  </svg>
))
```

src/client/components/dashboard/field/view.tsx
```diff
@@ -1,69 +1,52 @@
...
-  public componentWillUnmount() {
-    cancelAnimationFrame(this.rafId)
-  }
+  @action
+  private onResize = (width: number, height: number) => {
 
-  public render() {
-    return <canvas className={style.field} ref={this.onRef}/>
-  }
+    const m = this.props.model
```
BrendanAnnable: `const { model: { camera, fieldWidth, fieldLength } } = this.props;`

src/client/canvas/svg/marker.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { MarkerGeometry } from '../geometry/marker_geometry'
+import { Shape } from '../object/shape'
+
+import { svgAppearanceAttributes } from './svg'
+
+type Props = { model: Shape<MarkerGeometry>, world: Transform }
+export const Marker = observer(({ model: {
+  geometry: { x, y, radius, heading },
+  appearance,
+},                                world }: Props) => {
```
BrendanAnnable: This file needs autoformatting

src/client/canvas/svg/marker.tsx
```diff
@@ -0,0 +1,20 @@
...
+  const rotation = 135.0 + (180.0 / Math.PI) * Math.atan2(heading.y, heading.x)
+  return <path d='M-1 -1L0 -1A1 1 270 1 1 -1 0Z'
```
BrendanAnnable: I propose using a formatting like this (which I use elsewhere), simply because it makes the indentation look visually very similar to regular JavaScript:

```ts
return (
  <path
    d='M-1 -1L0 -1A1 1 270 1 1 -1 0Z'
    transform={`scale(${radius}) rotate(${rotation}) translate(${x}, ${y})`}
    {...svgAppearanceAttributes(appearance)}
  />
)
```

vs an imagined JavaScript version:

```ts
return {
  path: {
    d: '...',
    transform: '...',
    {...}
  }
};
```

src/client/components/dashboard/field/controller.ts
```diff
@@ -1,29 +0,0 @@
-import { action } from 'mobx'
-
-import { FieldModel } from './model'
-
-export class FieldController {
```
BrendanAnnable: Any reason for inlining the controller?

src/client/canvas/svg/svg.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const s = transform.scale
+  const r = (180.0 / Math.PI) * transform.rotate // SVG rotations are in degrees
+  const t = transform.translate
+  return `translate(${t.x}, ${t.y}) rotate(${r}) scale(${s.x}, ${s.y})`
+}
+
+type Props = { obj: Object2d, world: Transform }
+export function GeometryView({ obj, world }: Props): JSX.Element {
+
+  // TODO wrap things in appearance tags
```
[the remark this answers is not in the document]
TrentHouliston: It means I forgot to delete that comment!

src/client/canvas/renderer.ts
TrentHouliston: I had intended to either rename this or delete it, depending on how we felt about svg rendering vs canvas rendering

src/client/canvas/svg/text.tsx
```diff
@@ -0,0 +1,38 @@
...
+import * as React from 'react'
+import { Component } from 'react'
+
+import { Transform } from '../../math/transform'
+import { TextGeometry } from '../geometry/text_geometry'
+import { Shape } from '../object/shape'
+
+import { svgAppearanceAttributes, toSvgTransform } from './svg'
+
+
```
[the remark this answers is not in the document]
TrentHouliston: tslint allows up to 2 newlines in a row, not 3 or more

src/client/canvas/renderer.ts
BrendanAnnable: I don't think we're ready for deletion yet, especially since I'm not sure how it will perform vs SVG for more complex scenes. Chart will be an interesting test. I say we rename it for now.

src/client/components/dashboard/field/controller.ts
TrentHouliston: It didn't do anything anymore

src/client/canvas/svg/svg.tsx
```diff
@@ -0,0 +1,81 @@
...
+export function toSvgTransform(transform: Transform): string {
+  const s = transform.scale
+  const r = (180.0 / Math.PI) * transform.rotate // SVG rotations are in degrees
+  const t = transform.translate
+  return `translate(${t.x}, ${t.y}) rotate(${r}) scale(${s.x}, ${s.y})`
+}
+
+type Props = { obj: Object2d, world: Transform }
+
+export function GeometryView({ obj, world }: Props): JSX.Element {
```
BrendanAnnable: Can should also be marked as an `observer`

src/client/canvas/svg_renderer.tsx
```diff
@@ -0,0 +1,23 @@
+import { observer } from 'mobx-react'
+import * as React from 'react'
+import { Component } from 'react'
```
BrendanAnnable: unused

declarations/react-resize-detector.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module 'react-resize-detector' {
+  const value: any
```
BrendanAnnable:
```ts
declare module 'react-resize-detector' {
  import { ComponentType } from 'react'

  const component: ComponentType<{
    handleWidth?: boolean,
    handleHeight?: boolean,
    onResize: (width: number, height: number) => void
  }>

  export default component
}
```

src/client/canvas/svg/arc.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const { origin, radius, startAngle, endAngle, anticlockwise } = geometry
+
+  // Is this arc empty?
+  if (radius < 0) {
+    throw new Error(`Negative radius: ${radius}`)
+  }
+
+  // Calculate out our start and end points on the circle
+  const p0 = Vector2.fromPolar(radius, startAngle).add(origin)
+  const p1 = Vector2.fromPolar(radius, endAngle).add(origin)
```
BrendanAnnable: 👍  thanks

src/client/canvas/svg/tests/text.tests.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { Transform } from '../../../math/transform'
+import { BasicAppearance } from '../../appearance/basic_appearance'
+import { TextGeometry } from '../../geometry/text_geometry'
+import { Shape } from '../../object/shape'
+import { Text } from '../text'
+
+describe('TextSVGRenderer', () => {
+  it('renders', () => {
+    const model = Shape.of(TextGeometry.of(), BasicAppearance.of())
+    const world = Transform.of()
```
BrendanAnnable: Given text actually uses the world transform, can we add a non-identity transform here?

src/client/canvas/svg/text.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { TextGeometry } from '../geometry/text_geometry'
+import { Shape } from '../object/shape'
+
+import { toSvgProps, toSvgTransform } from './svg'
+
+type Props = { model: Shape<TextGeometry>, world: Transform }
+export const Text = observer(({ model: { geometry, appearance }, world }: Props) => {
+  const { x, y, fontFamily, text, textAlign, textBaseline, alignToView } = geometry
+
+  // TODO handle font size and max width properly
```
BrendanAnnable: Can you move this comment to the hardcoded `0.015` value, and maybe elaborate a bit. e.g.

`// TODO (Houliston): Calculate fontSize such that the text fits within the given maxWidth`

PR #153, opened 2018-01-14
Loosen member tslint restrictions

BrendanAnnable:
This PR

1. Removes member ordering restrictions, allowing a more flexible/sensible code structure.
2. Removes `public` accessors and assumes it to be the default. So much less typing to do!

Any merge conflicts with existing PRs should be easy to resolve by cherry picking `
7d0dd58` and running `yarn tslint:fix` before merging:

```
git cherry-pick -n 7d0dd58
yarn tslint:fix
git commit -a -m 'remove public accessors'
git fetch origin
git merge origin/master
```

### DIFF
--- a/src/client/canvas/appearance/basic_appearance.ts
+++ b/src/client/canvas/appearance/basic_appearance.ts
@@ -1,9 +1,9 @@
 import { observable } from 'mobx'
 
 export class BasicAppearance {
-  @observable public fillStyle: string
-  @observable public lineWidth: number
-  @observable public strokeStyle: string
+  @observable fillStyle: string
+  @observable lineWidth: number
+  @observable strokeStyle: string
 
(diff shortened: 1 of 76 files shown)


TrentHouliston:
I understand the removal of public, but why do we want to get rid of member ordering?

BrendanAnnable:
1. It doesn't have a fixer, so automatically on shaky ground.
2. It forces an ordering that does not always sense if you want to list your methods in a linear story such than you easily read a class. Sometimes you want short private methods (e.g. 1 line event handlers) right under the public methods that use them because it makes more sense than having all your completely unrelated private methods at the bottom. Basically the most logical choice isn't always the listing mandated by this tslint rule.

TrentHouliston:
Fair enough,
Glaring issues should get picked up in review anyway

PR #157, opened 2018-01-16
Refactor the canvas renderer into multiple files

TrentHouliston:
In anticipation of making a component that is equivalent to the SVGRenderer react component


src/client/render2d/canvas/util.ts
```diff
@@ -0,0 +1,79 @@
...
+      renderText(ctx, obj, world)
+    } else {
+      throw new Error(`Unsupported geometry type: ${obj.geometry}`)
+    }
+  } else {
+    throw new Error(`Unsupported Object2d type: ${obj}`)
+  }
+}
+
+export function applyTransform(ctx: CanvasRenderingContext2D, transform: Transform): void {
```
BrendanAnnable: Is this export needed?

src/client/render2d/canvas/util.ts
```diff
@@ -0,0 +1,79 @@
...
+import { renderCircle } from './circle'
+import { renderLine } from './line'
+import { renderMarker } from './marker'
+import { renderPolygon } from './polygon'
+import { renderText } from './text'
+
+export function renderObject2d(ctx: CanvasRenderingContext2D, obj: Object2d, world: Transform) {
+
+  if (obj instanceof Group) {
+    const newWorld = world.clone().then(obj.transform)
```
BrendanAnnable: `objWorld` perhaps?

src/client/render2d/canvas/text.ts
```diff
@@ -0,0 +1,37 @@
+import { Transform } from '../../math/transform'
+import { Vector2 } from '../../math/vector2'
+import { TextGeometry } from '../geometry/text_geometry'
+import { Shape } from '../object/shape'
+
+import { applyAppearance } from './util'
```
BrendanAnnable: How about `rendering` instead of `util`? Or some other word that describes the things that go in here.

src/client/render2d/canvas/util.ts
TrentHouliston: I thought it was... but apparently it's not

PR #158, opened 2018-01-16
Add sublime project

TrentHouliston:
(we're all secretly using it)


NUsight2.sublime-project
```diff
@@ -0,0 +1,21 @@
+{
+	"folders":
+	[
+        {
+            "folder_exclude_patterns":
+            [
+                "NUbots/build*",
+                ".vagrant"
+            ],
```
BrendanAnnable: What are these for?

BrendanAnnable:
you monsters

NUsight2.sublime-project
TrentHouliston: Oops...

NUsight2.sublime-project
```diff
@@ -0,0 +1,16 @@
+{
+	"folders":
+	[
+        {
+            "path": "."
```
BrendanAnnable: Surely this isn't still needed, is it needed? If so, can we keep the indentation consistent (2 spaces)?

TrentHouliston: path: "." is still needed for it to show up properly

PR #160, opened 2018-01-16
Stop SVG from making unneeded groups to reduce dom element count

TrentHouliston:
Most of the groups don't actually do anything other than nest svg elements in groups that transform nothing.


src/client/render2d/svg/group.tsx
```diff
@@ -7,8 +7,22 @@ import { Group as GroupGeometry } from '../object/group'
...
+      </>
+    )
+  } else {
+    return (
+      <g transform={toSvgTransform(transform)}>
+        {elems}
+      </g>
+    )
+  }
+})
```
BrendanAnnable: some ⛳️ ing, also pulled out the `objWorld` calculation since it is constant for the map call:

```ts
type Props = { model: GroupGeometry, world: Transform }
export const Group = observer(({ model: { children, transform }, world }: Props) => {
  const objWorld = transform.clone().then(world)
  const elems = children.map((obj, i) => (<GeometryView key={i} obj={obj} world={objWorld}/>))
  return transform.isIdentity()
    ? <>{elems}</> // If we have the identity transform forgo the group to save on dom elements
    : <g transform={toSvgTransform(transform)}>{elems}</g>
})
```

PR #162, opened 2018-01-17
[WIP] Chart Tree

JosephusPaye:
Work on the checkbox tree for the chart view, extracted from #127.


JosephusPaye:
Abandoned for #163. #163 is a simpler PR with only the checkbox-tree and none of the chart/line-chart code.

PR #163, opened 2018-01-17
Add checkbox tree

JosephusPaye:
Add a checkbox tree for use in the chart view. Based off  #162, which was extracted from #127.


src/client/components/checkbox_tree/tree_node/view.tsx
```diff
@@ -0,0 +1,129 @@
...
+export class TreeNode extends Component<TreeNodeProps> {
+  private checkbox: HTMLInputElement
+  private stopAutorun: IReactionDisposer
+
+  componentDidMount() {
+    if (!this.checkbox) {
+      return
+    }
+
+    this.stopAutorun = autorun(() => this.updateCheckbox())
```
TrentHouliston: Put a comment here to say why autorun is needed

src/client/components/checkbox_tree/tree_node/view.tsx
```diff
@@ -0,0 +1,131 @@
...
+    } else if (this.props.node.checked === CheckedState.Unchecked) {
+      this.checkbox.indeterminate = false
+      this.checkbox.checked = false
+    } else {
+      this.checkbox.indeterminate = true
+    }
+  }
+
+  private onCheckboxClick = (event: MouseEvent<HTMLInputElement>) => {
+    event.stopPropagation()
```
BrendanAnnable: Out of interest, why was this needed?

src/client/components/checkbox_tree/model.ts
```diff
@@ -0,0 +1,35 @@
...
+}
+
+export type TreeModelOpts = {
+  nodes: TreeNodeModel[]
+  usePessimisticToggle: boolean
+}
+
+export interface TreeModel {
+  nodes: TreeNodeModel[]
+  usePessimisticToggle: boolean
```
BrendanAnnable: Is this used somewhere?

src/client/components/checkbox_tree/model.ts
```diff
@@ -0,0 +1,35 @@
...
+  label: string
+  expanded: boolean
+}
+
+export interface TreeNodeModel {
+  children: TreeNodeModel[]
+  checked: CheckedState
+  label: string
+  expanded: boolean
+  leaf: boolean
```
BrendanAnnable: Is this used somewhere?

src/client/components/checkbox_tree/view.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+export const CheckboxTree: StatelessComponent<CheckboxTreeProps> = (props: CheckboxTreeProps) => {
+  return (
+    <div>
+      {
+        props.model.nodes.map((node, i) => {
+          return <TreeNode
+            node={node}
+            level={0}
+            key={i}
```
BrendanAnnable: To [prevent performance issues](https://reactjs.org/docs/lists-and-keys.html#keys), we might want to enforce that ever node has an `id` so that we could use `key={node.id}` here instead.

PR #165, opened 2018-01-26
Easy component installing

LukeFarrawell:
Use dynamic importing to allow for easy importing based on a JSON file.

Structure:
```JavaScript
{
  "default" : [] // a list of strings of the dependencies that is required by all modules
  "modules" : [
    {
      "name": "dashboard", // the component name as it appears in the folder structure
      "enable": true // should it be loaded in the client
      "dependencies": [] // a list of strings of all dependencies that are unique for this module
    },
  ]
}
```

### DIFF
--- a/components.json
+++ b/components.json
@@ -1,12 +1,12 @@
 {
   "modules" : [
-      {
-        "name": "dashboard",
-        "enable": true
-      },
-      {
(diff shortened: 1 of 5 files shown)


TrentHouliston:
Closed as we don't need it right now

PR #166, opened 2018-03-16
Upgrade dependencies

JosephusPaye:
This upgrades the dependencies with minor changes. Webpack and Typescript have been left for another PR.


package.json
```diff
@@ -100,22 +100,23 @@
...
+    "react-resize-detector": "^2.2.0",
     "react-router": "^4.2.0",
     "react-router-dom": "^4.2.2",
     "serve-favicon": "^2.4.3",
-    "socket.io": "^1.7.3",
-    "socket.io-client": "^1.7.3",
-    "three": "^0.84.0",
+    "socket.io": "^2.0.4",
+    "socket.io-client": "^2.0.4",
+    "three": "^0.91.0",
```
BrendanAnnable: The types are only for `0.89.12`.

package.json
```diff
@@ -100,22 +100,23 @@
...
-    "react-resize-detector": "^1.1.0",
+    "react-resize-detector": "^2.2.0",
     "react-router": "^4.2.0",
     "react-router-dom": "^4.2.2",
     "serve-favicon": "^2.4.3",
-    "socket.io": "^1.7.3",
-    "socket.io-client": "^1.7.3",
-    "three": "^0.84.0",
+    "socket.io": "^2.0.4",
+    "socket.io-client": "^2.0.4",
```
BrendanAnnable: The current types are only for `1.4.29`.

LukeFarrawell:
RIP I just went to decline, just missed, controls don't work.

package.json
```diff
@@ -100,22 +100,23 @@
...
+    "mobx": "^4.1.0",
+    "mobx-react": "^5.0.0",
+    "mobx-utils": "^4.0.0",
     "nuclearnet.js": "^1.5.2",
     "protobufjs": "6.7.3",
     "react": "^16.2.0",
     "react-color": "^2.13.8",
     "react-dom": "^16.0.0",
-    "react-resize-detector": "^1.1.0",
+    "react-resize-detector": "^2.2.0",
```
BrendanAnnable: Do our hand written types for this need to be updated?

TrentHouliston:
The changes for socket.io I tested and looked at. Most of the changes in 2.0 are not a change to the API but a change to the internals/network interfaces. So the types remaining behind is fine.

package.json
TrentHouliston: No, version 2.0 just swapped to using a native method (ResizeObserver) rather than the hacky scroll method.

package.json
TrentHouliston: Not a problem, the 2.0 was a change to the wire protocol etc rather than a change to the API (mainly) and the performance improvements are worth it.

TrentHouliston:
@LukeFarrawell You're right! I thought I tested everything but it seems that we missed something!

package.json
TrentHouliston: That's as high as they go :(
But I don't think we would use any of the bleeding edge features of ThreeJS. (not that we could with outdated types)

TrentHouliston: Seems this one might have been a problem though :P

TrentHouliston: Nevermind, it's fine we just accidentally deleted some code we shouldn't

TrentHouliston:
Turns out we accidentally removed the code that let the controls work when upgrading mobx.
Thought it was related to the autorun code.

package.json
TrentHouliston: Also the types for 2.0 don't exist yet

BrendanAnnable: There are definitely API changes that the old types won't let you use.

> performance improvements are worth it.

Have you tested that? I'm not sure simply upgrading to 2.0 has any wire protocol changes? IIRC you needed to explicitly import/use some new thing that the types don't know exists.

package.json
BrendanAnnable: > But I don't think we would use any of the bleeding edge features of ThreeJS.

That would indicate we should downgrade three.js to match the types.

JosephusPaye:
@LukeFarrawell See https://github.com/NUbots/NUsight2/pull/167.

package.json
TrentHouliston: I don’t think so, the changelog implied it changed the defaults. And merged the negotiation packets on the wire. The only breaking changes are irrelevant to us.

package.json
TrentHouliston: That is a valid point. We may have been overzealous with this PR :P

package.json
BrendanAnnable: Mmmk. 🤷‍♂️ I hope this doesn't cause someone a lot of confusion one day!

TrentHouliston: Well Socket.IO version 2.0 has been out for almost a year at this point... If the types haven't been updated in that time it implies that either nobody is using socket.io or there isn't anything to change. (or people are lazy)

PR #168, opened 2018-04-04
Add chart

JosephusPaye:
Adding chart, mostly adapted from @MonicaOlejniczak's #127 PR.


src/client/components/chart/line_chart/view.tsx
```diff
@@ -0,0 +1,69 @@
...
+  model: ChartModel
+}
+
+@observer
+export class LineChart extends Component<LineChartProps> {
+  render() {
+    const chartModel = this.props.model
+    const lineChartModel = LineChartModel.of(this.props.model)
+    const viewModel = LineChartViewModel.of(lineChartModel)
+    const controller = LineChartController.of(lineChartModel)
```
BrendanAnnable: The controller and model are dependencies that should be passed as props, not created in a `render` call.

Pushing a partial fix which moves the controller to the `install.tsx` and model up a level.

What is the expected lifetime of a `LineChartModel`? Shouldn't it be created at the same time as the `ChartModel`?

BrendanAnnable: Moved up `install.tsx` to match lifetime of `ChartModel`.

1be6469576e57c25f097c411af4efd694a284fa5

src/client/components/chart/model.ts
```diff
@@ -82,10 +82,7 @@ export class ChartModel {
   @computed
   get tree() {
     return {
-      nodes: Array.from(this.treeData.entries()).map((entry: [string, TreeData | DataSeries]) => TreeViewModel.of({
-        label: entry[0],
-        model: entry[1],
-      })),
+      nodes: Array.from(this.treeData.entries()).map(([label, model]) => TreeViewModel.of({ label, model })),
```
TrentHouliston: ⛳️
```typescript
nodes: Array.from(this.treeData.entries(), ([label, model]) => TreeViewModel.of({ label, model }))
```

BrendanAnnable: 🎉 e1b90ec5e055d2f5bd16523c106d0ed049885108

PR #170, opened 2018-04-10
Adds a pixijs based renderer for render2d

TrentHouliston:
Pixi uses WebGL as a backend, so for some workloads (e.g. chart) this may provide a benefit.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "@types/jest": "^22.0.1",
     "@types/minimist": "^1.2.0",
     "@types/node": "^9.4.7",
+    "@types/pixi.js": "^4.7.2",
     "@types/react": "^16.0.34",
     "@types/react-color": "^2.13.3",
     "@types/react-dom": "^16.0.2",
@@ -88,6 +89,7 @@
     "typescript": "^2.6.2",
     "url-loader": "0.6.2",
     "webpack": "^3.4.0",
+    "webpack-dashboard": "^1.1.1",
     "webpack-dev-middleware": "^2.0.3",
     "webpack-dev-server": "^2.6.1",
     "webpack-hot-middleware": "^2.20.0"
@@ -106,6 +108,7 @@
     "mobx-react": "^5.0.0",
     "mobx-utils": "^4.0.0",
     "nuclearnet.js": "^1.5.2",
+    "pixi.js": "^4.7.3",
     "protobufjs": "6.7.3",
     "react": "^16.2.0",
     "react-color": "^2.13.8",
--- a/package.json
+++ b/package.json
@@ -89,7 +89,6 @@
     "typescript": "^2.6.2",
     "url-loader": "0.6.2",
     "webpack": "^3.4.0",
-    "webpack-dashboard": "^1.1.1",
     "webpack-dev-middleware": "^2.0.3",
     "webpack-dev-server": "^2.6.1",
     "webpack-hot-middleware": "^2.20.0"
--- a/src/client/components/dashboard/dashboard_robot/model.ts
+++ b/src/client/components/dashboard/dashboard_robot/model.ts
@@ -100,7 +100,7 @@ export class DashboardRobotModel {
       robotPosition: Vector3.of(),
       robotPositionCovariance: Matrix3.of(),
       roleName: '',
-      textColor: '#fff',
+      textColor: '#ffffff',
       time: BrowserSystemClock.now(),
       voltage: -1,
       walkCommand: Vector3.of(),
--- a/src/client/components/dashboard/dashboard_robot/model.ts
+++ b/src/client/components/dashboard/dashboard_robot/model.ts
@@ -19,12 +19,13 @@ export class DashboardRobotModel {
 
   // Parameters that influence the display
   @observable camera: Transform
-  @observable ballColor: number
-  @observable ballSightColor: number
-  @observable kickTargetColor: number
+  @observable ballColor: string
+  @observable ballSightColor: string
+  @observable kickTargetColor: string
   @observable private robot: RobotModel
-  @observable robotColor: number
-  @observable textColor: number
+  @observable robotBorderColor: string
+  @observable robotColor: string
+  @observable textColor: string
 
   // Parameters from the network
   // The timestamp of the last message from the robot (in seconds since an arbitrary time)
@@ -78,10 +79,10 @@ export class DashboardRobotModel {
 
   static of = memoize((robot: RobotModel): DashboardRobotModel => {
     return new DashboardRobotModel(robot, {
-      ballColor: 0xff9800,
+      ballColor: '#ff9800',
       ballCovariance: Matrix2.of(),
       ballPosition: Vector2.of(),
-      ballSightColor: 0x4DB6AC,
+      ballSightColor: '#4DB6AC',
       battery: -1,
       behaviourState: State.UNKNOWN,
       camera: Transform.of(),
@@ -89,16 +90,17 @@ export class DashboardRobotModel {
       gamePhase: Phase.UNKNOWN_PHASE,
       playerId: -1,
       kickTarget: Vector2.of(),
-      kickTargetColor: 0x00796B,
+      kickTargetColor: '#00796B',
       lastCameraImage: 0,
       lastSeenBall: 0,
       lastSeenGoal: 0,
       penaltyReason: PenaltyReason.UNKNOWN_PENALTY_REASON,
-      robotColor: 0x015457,
+      robotBorderColor: 'transparent',
+      robotColor: '#015457',
       robotPosition: Vector3.of(),
       robotPositionCovariance: Matrix3.of(),
       roleName: '',
-      textColor: 0xffffff,
+      textColor: '#fff',
       time: BrowserSystemClock.now(),
       voltage: -1,
       walkCommand: Vector3.of(),
@@ -129,11 +131,12 @@ export class DashboardRobotModel {
 
 interface DashboardRobotModelOpts {
   camera: Transform
-  ballColor: number
-  ballSightColor: number
-  kickTargetColor: number
-  robotColor: number
-  textColor: number
+  ballColor: string
+  ballSightColor: string
+  kickTargetColor: string
+  robotBorderColor: string
+  robotColor: string
+  textColor: string
   time: number
   roleName: string
   battery: number
--- a/src/client/components/dashboard/dashboard_robot/model.ts
+++ b/src/client/components/dashboard/dashboard_robot/model.ts
@@ -19,13 +19,12 @@ export class DashboardRobotModel {
 
   // Parameters that influence the display
   @observable camera: Transform
-  @observable ballColor: string
-  @observable ballSightColor: string
-  @observable kickTargetColor: string
+  @observable ballColor: number
+  @observable ballSightColor: number
+  @observable kickTargetColor: number
   @observable private robot: RobotModel
-  @observable robotBorderColor: string
-  @observable robotColor: string
-  @observable textColor: string
+  @observable robotColor: number
+  @observable textColor: number
 
   // Parameters from the network
   // The timestamp of the last message from the robot (in seconds since an arbitrary time)
@@ -79,10 +78,10 @@ export class DashboardRobotModel {
 
   static of = memoize((robot: RobotModel): DashboardRobotModel => {
     return new DashboardRobotModel(robot, {
-      ballColor: '#ff9800',
+      ballColor: 0xff9800,
       ballCovariance: Matrix2.of(),
       ballPosition: Vector2.of(),
-      ballSightColor: '#4DB6AC',
+      ballSightColor: 0x4DB6AC,
       battery: -1,
       behaviourState: State.UNKNOWN,
       camera: Transform.of(),
@@ -90,17 +89,16 @@ export class DashboardRobotModel {
       gamePhase: Phase.UNKNOWN_PHASE,
       playerId: -1,
       kickTarget: Vector2.of(),
-      kickTargetColor: '#00796B',
+      kickTargetColor: 0x00796B,
       lastCameraImage: 0,
       lastSeenBall: 0,
       lastSeenGoal: 0,
       penaltyReason: PenaltyReason.UNKNOWN_PENALTY_REASON,
-      robotBorderColor: 'transparent',
-      robotColor: '#015457',
+      robotColor: 0x015457,
       robotPosition: Vector3.of(),
       robotPositionCovariance: Matrix3.of(),
       roleName: '',
-      textColor: '#fff',
+      textColor: 0xffffff,
       time: BrowserSystemClock.now(),
       voltage: -1,
       walkCommand: Vector3.of(),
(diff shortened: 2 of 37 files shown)


package.json
```diff
@@ -88,6 +89,7 @@
     "typescript": "^2.6.2",
     "url-loader": "0.6.2",
     "webpack": "^3.4.0",
+    "webpack-dashboard": "^1.1.1",
```
BrendanAnnable: 🙅‍♂️

src/client/components/dashboard/dashboard_robot/model.ts
```diff
@@ -79,28 +78,27 @@ export class DashboardRobotModel {
 
   static of = memoize((robot: RobotModel): DashboardRobotModel => {
     return new DashboardRobotModel(robot, {
-      ballColor: '#ff9800',
+      ballColor: 0xff9800,
       ballCovariance: Matrix2.of(),
       ballPosition: Vector2.of(),
-      ballSightColor: '#4DB6AC',
+      ballSightColor: 0x4DB6AC,
```
BrendanAnnable: Seeing the raw number `5093036` in the console as a variable is going to suck, the bit-shifting needed below is already a bad sign. Can we go the other way and parse hex strings in the pixi renderer? `parseInt('#4DB6AC'.slice(1), 16)` reads much nicer.

src/client/render2d/pixi/rendering.ts
```diff
@@ -0,0 +1,91 @@
...
+    exec(obj)
+    obj.endFill()
+  } else if (appearance instanceof LineAppearance) {
+    obj.lineStyle(appearance.lineWidth, appearance.strokeColor, appearance.strokeAlpha)
+    exec(obj)
+    // ctx.lineCap = appearance.lineCap
+    // ctx.lineDashOffset = appearance.lineDashOffset
+    // ctx.lineJoin = appearance.lineJoin
+    // ctx.lineWidth = appearance.lineWidth
+    // ctx.strokeStyle = appearance.strokeStyle
```
BrendanAnnable: remove?

src/client/render2d/pixi_renderer.tsx
```diff
@@ -0,0 +1,114 @@
...
+    cam.y = transform.translate.y
+    cam.scale.x = transform.scale.x
+    cam.scale.y = transform.scale.y
+    cam.addChild(renderObject2d(this.props.scene))
+
+    this.renderer.render(cam)
+  }
+
+  @action
+  private onResize = (width: number, height: number) => {
```
BrendanAnnable: Can we avoid copy/pasting this?

BrendanAnnable: Actually this looks slightly different, why is that?

package.json
TrentHouliston: How did that get there!

src/client/components/dashboard/dashboard_robot/model.ts
TrentHouliston: I thought about it, but it ends up being dangerous since before it was just a style which could have anything in it. Limiting it down to colour works better. Is there some typescript way that I can enforce it to be a string of the form `#[A-Za-z0-9]{6}`

src/client/render2d/pixi_renderer.tsx
TrentHouliston: The renderer has a viewport etc in it that needs adjusting

src/client/render2d/pixi/rendering.ts
BrendanAnnable: Oh, these are TODOs?

TrentHouliston: Yeah pixi doesn't have a way to do them in their API so i'm leaving them there at the moment.
What do you think I should do with them?

src/client/components/dashboard/dashboard_robot/model.ts
BrendanAnnable: Just validate it at parse time, I think that should be fine and the least annoying.

```ts
function hexColorInt(hex: string): number {
  if (!hex.match(/^#[0-9a-f]{6}$/) {
    throw new Error(`Invalid color: ${hex}`)
  }
  return parseInt('#4DB6AC'.slice(1), 16)
}
```

src/client/render2d/pixi/rendering.ts
```diff
@@ -0,0 +1,110 @@
...
+      return renderText(obj)
+    } else {
+      throw new Error(`Unsupported geometry type: ${obj.geometry}`)
+    }
+  } else {
+    throw new Error(`Unsupported Object2d type: ${obj}`)
+  }
+})
+
+const toPixiColor = (style: string): [number, number] => {
```
BrendanAnnable: Can we return an object, so that the values are clear at the usage site? and add a comment describing the range: e.g.
```ts
{
  color: number, // range: [0, 0xFFFFFF]
  alpha: number, // range: [0, 1]
}
```

src/client/render2d/pixi/rendering.ts
```diff
@@ -0,0 +1,110 @@
...
+  if (style === 'transparent') {
+    return [0, 0]
+  }
+  let result = /^#([A-Fa-f0-9])([A-Fa-f0-9])([A-Fa-f0-9])$/.exec(style)
```
BrendanAnnable: What does this do that the the version below does not? Can we add a comment?

src/client/render2d/pixi/rendering.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+  result = /^#([A-Fa-f0-9]{6})$/.exec(style)
+  if (result !== null) {
+    return [parseInt(result[1], 16), 1]
+  }
+
+  throw new Error('Pixi cannot handle non hex colours')
+}
+
+export function applyAppearance(obj: Graphics, appearance: Appearance, exec: (obj: Graphics) => void): void {
```
BrendanAnnable: Can we rename `exec` to `draw`?

src/client/render2d/pixi/rendering.ts
BrendanAnnable: Can we either throw e.g. `throw new Error('lineCap not supported by pixi engine')` or add explicit TODOs
```ts
// TODO: Support lineCap, lineCap, lineJoin, lineWidth, strokeStyle
```

src/client/render2d/pixi/text.ts
```diff
@@ -0,0 +1,37 @@
...
+      fontFamily,
+      padding: 10,
+      textBaseline: 'middle',
+      fill: shape.appearance.fillStyle,
+      align: textAlign,
+    })
+    t.x = x
+    t.y = y
+    t.anchor.x = 0.5
+    t.anchor.y = 0.75 // Might be hacks? might not be...
```
BrendanAnnable: What does this mean 👀

src/client/render2d/pixi/rendering.ts
```diff
@@ -0,0 +1,110 @@
...
+export const renderObject2d = createTransformer((obj: Object2d): DisplayObject => {
+
+  if (obj instanceof Group) {
+    const g = new Container()
+
+    g.scale.x = obj.transform.scale.x
+    g.scale.y = obj.transform.scale.x
+    g.x = obj.transform.translate.x
+    g.y = obj.transform.translate.y
+    g.rotation = obj.transform.rotate
```
BrendanAnnable: ⛳️ 
```ts
const { transform: { scale, translate, rotate } } = obj;
g.scale.x = scale.x
g.scale.y = scale.y
...
```

src/client/render2d/pixi_renderer.tsx
BrendanAnnable: Can we add a jsdoc comment to this method describing what it is trying to achieve, the comments below only say how it achieves that, but it isn't clear what its trying to do.

src/client/render2d/pixi/rendering.ts
BrendanAnnable: Right, I wouldn't have an issue dropping support for the 3 letter version (like we did with `yellow`/`orange` etc) so that all our colours are consistent. Less things to support and I don't think anyone will care.

src/client/render2d/pixi_renderer.tsx
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    this.renderer = autoDetectRenderer({
+      view: this.canvas,
+      transparent: true,
+      antialias: true,
+    })
+
+    // Render when changes happen
+    this.stopAutorun = autorun(() => this.renderCanvas(), {
```
BrendanAnnable: unnecessary lambda here
```ts
this.stopAutorun = autorun(this.renderCanvas, {
...
})
```

src/client/render2d/pixi_renderer.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+    const cam = new Container()
+    cam.rotation = transform.rotate
+    cam.x = transform.translate.x
+    cam.y = transform.translate.y
+    cam.scale.x = transform.scale.x
+    cam.scale.y = transform.scale.y
+    cam.addChild(renderObject2d(this.props.scene))
+
+    this.renderer.render(cam)
```
BrendanAnnable: I'm confused what's going on here, does the `renderObject2d` call above do the rendering? or does this line?

TrentHouliston: renderObject2d is the transformer that transforms the scene graph into a pixi scene graph. The function was taken from canvas and probably would have a better name in this case

src/client/render2d/pixi_renderer.tsx
TrentHouliston: This has been done on every single autorun call throughout the codebase.
Should I change them all?

TrentHouliston: Seems it's important for the localisation one at least, the other two work with this change

src/client/render2d/pixi_renderer.tsx
BrendanAnnable: Right, yeah in normal 2d canvas land side-effects happen along the way, and hence `render` is a good prefix. It looks like pixi takes a scene graph similar to three.js and the side-effects happen later, so it doesn't actually do rendering here.

I call this the view model in three.js land. Perhaps here it should be named something like `toPixiScene`?

src/client/render2d/pixi/rendering.ts
```diff
@@ -0,0 +1,110 @@
...
+      return renderText(obj)
+    } else {
+      throw new Error(`Unsupported geometry type: ${obj.geometry}`)
+    }
+  } else {
+    throw new Error(`Unsupported Object2d type: ${obj}`)
+  }
+})
+
+const toPixiColor = (style: string): {color: number, alpha: number} => {
```
BrendanAnnable: Add previously suggested output range comment please 🙏 `number` isn't sufficient.

src/client/render2d/pixi_renderer.tsx
```diff
@@ -0,0 +1,119 @@
...
+    cam.scale.y = transform.scale.y
+    cam.addChild(pixiObject(this.props.scene))
+
+    this.renderer.render(cam)
+  }
+
+  /**
+   * This function is executed when the canvas resizes.
+   * It is responsible for updating the viewport used by pixi
+   * as well as adjusting the scale of the world to normalize it
```
BrendanAnnable: > as well as adjusting the scale of the world to normalize it

I'm still not sure what this means, can it describe what effects/benefits this will have?

src/client/components/dashboard/field/view.tsx
```diff
@@ -20,7 +19,8 @@ export class Field extends Component<FieldProps> {
     const model = this.props.model
     const viewModel = FieldViewModel.of(model)
     return <div className={style.container}>
-      <SVGRenderer
+      <Renderer
+        engine='pixi'
```
BrendanAnnable: Revert?

src/client/render2d/pixi/rendering.ts
```diff
@@ -0,0 +1,110 @@
...
+  // Colors of the form #FFFFFF
+  result = /^#([A-Fa-f0-9]{6})$/.exec(style)
+  if (result !== null) {
+    return { color: parseInt(result[1], 16), alpha: 1 }
+  }
+
+  throw new Error('Pixi cannot handle non hex colours')
+}
+
+export function applyAppearance(obj: Graphics, appearance: Appearance, draw: (obj: Graphics) => void): void {
```
BrendanAnnable: I'd like to explore how we can remove the need for the `draw` callback, but that is for another PR I think.

src/client/render2d/pixi_renderer.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+@observer
+export class PixiRenderer extends Component<RendererProps> {
+  @observable private resolution: Transform = Transform.of()
+  private canvas: HTMLCanvasElement
+  private renderer: WebGLRenderer | CanvasRenderer
+  private stopAutorun: IReactionDisposer
+
+  componentDidMount() {
+    if (!this.canvas) {
```
BrendanAnnable: I'd prefer this to error at runtime, because it shouldn't happen. Just use `this.canvas!` below.

src/client/render2d/pixi_renderer.tsx
```diff
@@ -0,0 +1,119 @@
...
+import { Transform } from '../math/transform'
+
+import { pixiObject } from './pixi/rendering'
+import { RendererProps } from './renderer_props'
+import * as style from './style.css'
+
+@observer
+export class PixiRenderer extends Component<RendererProps> {
+  @observable private resolution: Transform = Transform.of()
+  private canvas: HTMLCanvasElement
```
BrendanAnnable: This should be `private canvas: HTMLCanvasElement | null = null`

src/client/render2d/pixi_renderer.tsx
```diff
@@ -0,0 +1,119 @@
...
+          className={style.container}
+          width={-this.resolution.translate.x * 2}
+          height={-this.resolution.translate.y * 2}
+          ref={this.onRef}
+        />
+      </div>
+    )
+  }
+
+  private onRef = (canvas: HTMLCanvasElement) => {
```
BrendanAnnable: This should be `canvas: HTMLCanvasElement | null`, I'm not sure how its passing the build without this.

src/client/render2d/pixi/rendering.ts
TrentHouliston: I don't 100% understand what you mean

src/client/render2d/pixi_renderer.tsx
TrentHouliston: Added more comments

src/client/components/dashboard/field/view.tsx
TrentHouliston: swapped back to svg

src/client/render2d/pixi/rendering.ts
BrendanAnnable:
```ts
const toPixiColor = (style: string): {
  color: number, // range: [0, 0xFFFFFF]
  alpha: number, // range: [0, 1]
} => {
...
}
```

src/client/render2d/pixi/rendering.ts
TrentHouliston: The draw callback is used to wrap the drawing code with an appearance.
It handles code that needs to be executed before/after drawing to make it look how we want.

PR #171, opened 2018-04-11
Refactor the Appearance class

TrentHouliston:
As more rendering engines have been implemented the appearance class has gotten harder to implement.
This refactors it so it is easier to use.

### DIFF
--- a/src/client/render2d/canvas/rendering.ts
+++ b/src/client/render2d/canvas/rendering.ts
@@ -84,7 +84,7 @@ export function applyAppearance(ctx: CanvasRenderingContext2D, appearance: Appea
 
   if (appearance instanceof BasicAppearance) {
 
-    if(appearance.fill) {
+    if (appearance.fill) {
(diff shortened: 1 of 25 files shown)


src/client/render2d/canvas/rendering.ts
```diff
@@ -65,38 +65,42 @@ export function applyTransform(ctx: CanvasRenderingContext2D, transform: Transfo
   ctx.rotate(transform.rotate * (transform.anticlockwise ? 1 : -1))
 }
 
-const hexToRGB = (hex: string): [number, number, number] => {
+export const hexToRGB = (hex: string): { r: number, g: number, b: number} => {
   const result = /^#([A-Fa-f0-9]{2})([A-Fa-f0-9]{2})([A-Fa-f0-9]{2})/.exec(hex)
 
   if (result === null) {
     throw Error(`Color ${hex} was not a hex color`)
```
BrendanAnnable: test this case too 😄

PR #173, opened 2018-04-16
Update the protocol buffers and the protobufjs library

TrentHouliston:
The API changed slightly for protobufjs, this implements the new one as well as updating the protocol buffers from NUbots


package.json
```diff
@@ -17,6 +17,8 @@
     "tslint:fix": "tslint -c tslint.json 'src/**/*.ts{,x}' --exclude 'src/shared/proto/**' --fix",
     "tscheck": "tsc --project tsconfig.json",
     "tscheck:watch": "yarn tsc --watch",
+    "message:js": "$(npm bin)/pbjs -t static-module -w commonjs -o src/shared/proto/messages.js -p src/shared/proto $(find src/shared/proto -name \"*.proto\")",
```
JosephusPaye: You can remove `$(npm bin)/pbjs` for `pbjs` as yarn/npm will try to resolve the command from `node_modules/.bin/`.

As you know, it doesn't work with the `$()` on Windows (no Bash).

This makes `$(find src/shared/proto -name \"*.proto\")` fail as well. An alternative may be doing the generation grammatically. There's a [command line api](https://github.com/dcodeIO/protobuf.js#command-line-api) for protobuf.js that can be used.

PR #176, opened 2018-04-19
Implement basic vision view

TrentHouliston:
Implements a basic vision view.
Shows each robot's camera streams


src/client/components/vision/network.ts
```diff
@@ -0,0 +1,50 @@
...
+    const robot = VisionRobotModel.of(robotModel)
+
+    if (!robot.cameras.has(image.cameraId)) {
+      robot.cameras.set(image.cameraId, CameraModel.of({
+        model: robot,
+        id: image.cameraId,
+      }))
+    }
+
+    const camera = robot.cameras.get(image.cameraId)!
```
BrendanAnnable: ⛳️ to both remove the `!` here and make use of destructuring to reduce chains.

```ts
  @action
  private onImage = (robotModel: RobotModel, image: Image) => {
    const robot = VisionRobotModel.of(robotModel)
    const { cameraId, name, dimensions, format, data, Hcw } = image;

    let camera = robot.cameras.get(cameraId)
    if (!camera) {
      camera = CameraModel.of({ model: robot, id: cameraId })
      robot.cameras.set(cameraId, camera)
    }

    camera.image = {
      width: dimensions!.x!,
      height: dimensions!.y!,
      format: format!,
      data,
      Hcw: Matrix4.from(Hcw),
    }
    camera.name = name
  }
```

src/client/components/vision/camera/model.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import { Matrix4 } from '../../../math/matrix4'
+import { VisionRobotModel } from '../model'
+
+export class ImageModel {
+  @observable width: number = 1280
+  @observable height: number = 1024
+  @observable format: number = 0
+  @observable.ref data: Uint8Array
+  @observable Hcw: Matrix4 = Matrix4.of()
```
BrendanAnnable: Defaults in factories.

src/client/components/vision/camera/model.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export class CameraModel {
+
+  private model: VisionRobotModel
+  readonly id: number
+
+  @observable image?: ImageModel
+  @observable name: string = ''
```
BrendanAnnable: Defaults in factories.

src/client/components/vision/camera/model.ts
```diff
@@ -0,0 +1,32 @@
...
+  constructor(model: VisionRobotModel, id: number) {
+    this.model = model
+    this.id = id
+  }
+
+  static of(opts: {
+    model: VisionRobotModel
+    id: number}) {
+    return new CameraModel(opts.model, opts.id)
+  }
```
BrendanAnnable: ⛳️ 
```ts
  static of({ model, id }: { model: VisionRobotModel, id: number }) {
    return new CameraModel(model, id)
  }
```

src/client/components/vision/camera/view.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { Component } from 'react'
+
+import { CameraViewModel } from './view_model'
+
+@observer
+export class CameraView extends Component<{ viewModel: CameraViewModel }> {
+  private destroy: () => void = () => {}
+
+  componentDidMount() {
+    this.destroy = autorun(() => this.renderScene(), {
```
BrendanAnnable: ⛳️ 
```ts
this.destroy = autorun(this.renderScene, { scheduler: requestAnimationFrame })
```

Then use method binding below:

```ts
private renderScene = () => {
```

src/client/components/vision/camera/view.tsx
```diff
@@ -0,0 +1,61 @@
...
+  render() {
+
+    const { width, height } = this.props.viewModel
+
+    const aspectRatio = width / height
+    const percentage = 60
+
+    return (
+      <canvas style={{
+        display: 'inline',
```
BrendanAnnable: Move non-computed styling to CSS classes, but I'm not sure what you are looking to achieve with this style anyway?

I suspect its for keeping camera views next to each other, in which case the parent should be responsible for that (e.g. with `display: flex`).

src/client/components/vision/camera/view.tsx
```diff
@@ -0,0 +1,61 @@
...
+      }}
+        width={width}
+        height={height}
+        ref={this.onRef}/>
+    )
+  }
+
+  @action
+  private onRef = (canvas: HTMLCanvasElement | null) => {
+    this.props.viewModel.canvas = canvas
```
BrendanAnnable: FYI I will be battling to remove this, but not in this PR.

src/client/components/vision/camera/view.tsx
```diff
@@ -0,0 +1,61 @@
...
+  private renderScene() {
+    const { viewModel } = this.props
+    const { canvas, scene, camera } = viewModel
+    const renderer = viewModel.renderer(canvas)
+    if (renderer) {
+      renderer.render(scene, camera)
+
+      // This is needed as we constantly make new scene objects, if we don't do this the renderLists
+      // stores a reference to the scene and mesh causing a memory leak
+      renderer.renderLists.dispose()
```
BrendanAnnable: Do we need all the other dispose methods if we use this?

I'd like to know more about this, is this a bug/issue with three.js? Is there a ticket or three.js source code you could link to?

src/client/components/vision/camera/view_model.ts
```diff
@@ -0,0 +1,282 @@
...
+import { Geometry } from 'three'
+
+import { memoize } from '../../../base/memoize'
+
+import { CameraModel } from './model'
+import { ImageModel } from './model'
+import * as bayerFragmentShader from './shaders/bayer.frag'
+import * as bayerVertexShader from './shaders/bayer.vert'
+
+export function fourcc(code: string): number {
```
BrendanAnnable: What does this do? Either document through comments or tests.

src/client/components/vision/camera/view_model.ts
```diff
@@ -0,0 +1,282 @@
...
+export class CameraViewModel {
+  @observable.ref canvas: HTMLCanvasElement | null = null
+
+  constructor(private model: CameraModel) {
+  }
+
+  // See this createTransformer here? that means that once it is no longer observed
+  // it will be deleted. However the model that creates it will not.
+  // This means that any @computed fields will only live so long as the view does as
+  // this viewModel will only live so long as the view does.
```
BrendanAnnable: This comment seems like generic mobx documentation, which indicates this isn't the place to put it.

I think we can either remove or add it to a wiki/readme somewhere.

src/client/components/vision/camera/view_model.ts
```diff
@@ -0,0 +1,282 @@
...
+    const scene = new Scene()
+    if (this.model.image) {
+      scene.add(this.image(this.model.image))
+    }
+    return scene
+  }
+
+  @computed
+  get width() {
+    return this.model.image ? this.model.image.width : 1280
```
BrendanAnnable: This default is basically arbitrary.

I think it would be better to tell the truth and return `number | undefined` and let the view decide what to do.

src/client/components/vision/camera/view_model.ts
```diff
@@ -0,0 +1,282 @@
...
+  }
+
+  @computed
+  get height() {
+    return this.model.image ? this.model.image.height : 1024
```
BrendanAnnable: ditto

src/client/components/vision/camera/view_model.ts
```diff
@@ -0,0 +1,282 @@
...
+  }
+
+  private image = createTransformer((image: ImageModel): Mesh => {
+    const mesh =  new Mesh(this.quadGeometry, this.imageMaterial(image))
+    mesh.scale.y = -1 // We need to flip Y to fix the image coordinates
```
BrendanAnnable: What does `fix` mean? What is wrong with the image coordinates? Can we move any flipping to the texture loading instead?

src/client/components/vision/camera/view_model.ts
```diff
@@ -0,0 +1,282 @@
...
+        return this.bayerTexture(image).texture
+      case fourcc('RGB3'):
+        return this.rgbTexture(image)
+      case fourcc('YUYV'):
+        return this.yuyvTexture(image)
+      case fourcc('GREY'):
+      case fourcc('GRAY'):
+        return this.grayTexture(image)
+      default:
+        throw Error('Unsupported image format')
```
BrendanAnnable:
```ts
throw Error(`Unsupported image format: ${image.format}`)
```

src/client/components/vision/camera/view_model.ts
```diff
@@ -0,0 +1,282 @@
...
+    switch (image.format) {
+      case fourcc('GRBG'):
+      case fourcc('RGGB'):
+      case fourcc('GBRG'):
+      case fourcc('BGGR'):
+        return this.bayerTexture(image).texture
+      case fourcc('RGB3'):
+        return this.rgbTexture(image)
+      case fourcc('YUYV'):
+        return this.yuyvTexture(image)
```
BrendanAnnable: Can we just delete this case and let the unsupported error happen?

src/client/components/vision/camera/view_model.ts
```diff
@@ -0,0 +1,282 @@
...
+
+    const { width, height } = image
+    const renderTarget = new WebGLRenderTarget(width, height)
+    renderTarget.depthBuffer = false
+    renderTarget.stencilBuffer = false
+    const scene = new Scene()
+    const material = this.bayerShader
+
+    // TODO: This is wrong and not very reactive however the alternative is recompiling the shader every time
+    // you have a better idea?
```
BrendanAnnable: Why is it recompiling the shader each time exactly?

src/client/components/vision/camera/view_model.ts
```diff
@@ -0,0 +1,282 @@
...
+import { ClampToEdgeWrapping } from 'three'
+import { LinearFilter } from 'three'
+import { NearestFilter } from 'three'
+import { PlaneGeometry } from 'three'
+import { Matrix4 } from 'three'
+import { Vector4 } from 'three'
+import { Vector2 } from 'three'
+import { Geometry } from 'three'
+
+import { memoize } from '../../../base/memoize'
```
BrendanAnnable: unused, any quite a few more above

package.json
```diff
@@ -53,7 +55,7 @@
     "@types/serve-favicon": "^2.2.28",
     "@types/socket.io": "^1.4.29",
     "@types/socket.io-client": "^1.4.29",
-    "@types/three": "^0.89.12",
+    "@types/three": "^0.91.0",
```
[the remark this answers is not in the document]
TrentHouliston: They did!

src/client/components/vision/camera/view_model.ts
TrentHouliston: Every time you make a material, it compiles the shader needed for that material. Previously the Bayer material was made here but because of that, it recompiled the shader every time.

src/shared/base/random/seeded_random.ts
```diff
@@ -13,12 +13,12 @@ export class SeededRandom {
     return new SeededRandom(seedrandom(seed))
   }
 
-  float(): number {
-    return this.prng()
+  float(min: number = 0, max: number = min + 1): number {
```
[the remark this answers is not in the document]
TrentHouliston: Reverted

src/client/components/vision/model.ts
```diff
@@ -0,0 +1,52 @@
+import { computed } from 'mobx'
+import { observable } from 'mobx'
+
+import { memoize } from '../../base/memoize'
+import { Matrix4 } from '../../math/matrix4'
+import { Vector2 } from '../../math/vector2'
+import { Vector3 } from '../../math/vector3'
```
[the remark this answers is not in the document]
TrentHouliston: 👍

src/client/components/vision/view_model.ts
```diff
@@ -0,0 +1,51 @@
...
+  @computed
+  get name() {
+    return this.model.name
+  }
+
+  @computed
+  get cameras(): CameraViewModel[] {
+    const cams = Array.from(this.model.cameras.values()).map(cam => CameraViewModel.of(cam))
+    cams.sort((a, b) => a.id - b.id)
+    return cams
```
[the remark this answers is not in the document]
TrentHouliston: 👍

src/client/components/vision/view_model.ts
```diff
@@ -0,0 +1,51 @@
...
+  constructor(private model: VisionModel) {
+  }
+
+  static of = createTransformer((model: VisionModel): VisionViewModel => {
+    return new VisionViewModel(model)
+  })
+
+  @computed
+  get robots(): RobotViewModel[] {
+    return this.visibleRobots.map(robot => RobotViewModel.of(robot))
```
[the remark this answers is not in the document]
TrentHouliston: 👍

src/client/components/vision/view.tsx
```diff
@@ -0,0 +1,44 @@
+import { autorun } from 'mobx'
```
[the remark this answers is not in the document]
TrentHouliston: 👍

src/client/components/vision/view.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { observer } from 'mobx-react'
+import * as React from 'react'
+import { Component } from 'react'
+import { ComponentType } from 'react'
+
+import { CameraView } from './camera/view'
+import { CameraViewModel } from './camera/view_model'
```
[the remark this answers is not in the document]
TrentHouliston: 👍

src/client/components/vision/view.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { VisionNetwork } from './network'
+import * as styles from './styles.css'
+import { VisionViewModel } from './view_model'
+import { RobotViewModel } from './view_model'
```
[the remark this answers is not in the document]
TrentHouliston: 👍

src/client/components/vision/view.tsx
```diff
@@ -0,0 +1,44 @@
...
+  viewModel: VisionViewModel
+  network: VisionNetwork
+  Menu: ComponentType
+}> {
+  componentWillUnmount() {
+    this.props.network.destroy()
+  }
+
+  render() {
+    const { viewModel, Menu } = this.props
```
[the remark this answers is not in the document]
TrentHouliston: 👍

src/client/components/vision/model.ts
```diff
@@ -0,0 +1,52 @@
...
+  constructor(private appModel: AppModel) {
+  }
+
+  static of = memoize((appModel: AppModel) => {
+    return new VisionModel(appModel)
+  })
+
+  @computed
+  get robots(): VisionRobotModel[] {
+    return this.appModel.robots.map(robot => VisionRobotModel.of(robot))
```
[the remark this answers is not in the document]
TrentHouliston: 👍

src/client/components/vision/network.ts
TrentHouliston: 👍

src/client/components/vision/camera/model.ts
TrentHouliston: 👍

src/client/components/vision/camera/model.ts
TrentHouliston: 👍

src/client/components/vision/network.ts
```diff
@@ -0,0 +1,50 @@
+import { action } from 'mobx'
+
+import { message } from '../../../shared/proto/messages'
+import { Matrix4 } from '../../math/matrix4'
+import { Vector3 } from '../../math/vector3'
```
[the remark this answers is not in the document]
TrentHouliston: 👍

src/client/components/vision/camera/model.ts
TrentHouliston: 👍

src/client/components/vision/camera/view.tsx
TrentHouliston: 👍

src/client/components/vision/camera/view.tsx
TrentHouliston: 👍

src/client/components/vision/camera/view.tsx
TrentHouliston: I'm not 100% sure, and honestly this one is only here as it's the only way to dispose of scenes and cameras that are no longer used.
I'm actually trying to think of a good way to reuse the same scene object each time so I don't have to resort to this sledgehammer.

src/client/components/vision/camera/view.tsx
TrentHouliston: Good luck :P, need to get that renderer somehow

src/client/components/vision/camera/view_model.ts
TrentHouliston: 👍

src/client/components/vision/camera/view_model.ts
TrentHouliston: I'm trying to explain that even though the @computed fields don't have a disposer they are ok since they are attached to the viewmodel/view. And when the viewmodel/view dies they will also be cleaned up.

src/client/components/vision/camera/view_model.ts
TrentHouliston: The main reason I did this is that if a camera image isn't being sent later on, but other objects like balls/goals are, it's going to freak out if it doesn't know what size to render with.
There may be a better way to handle this in the future, but for now since we only handle images i'll remove it.

src/client/components/vision/camera/view_model.ts
TrentHouliston: I'll change the comment to say what I meant to say :P

TrentHouliston: On second thought it's really not needed at all

TrentHouliston:
Testing is hard without doing full integration tests or otherwise doing view testing.
It might be possible to implement a few basic tests that ensure that the renderer works for render targets, but I'm not sure they test anything that's not blatantly obvious if it failed.
As for refactoring having this as a baseline for performance should let you refactor away and so long as the performance stays and no new memory leaks are caused it should be fine.

PR #182, opened 2018-04-26
Draw vision world lines

TrentHouliston:
Draws lines in world space using a shader. It is able to draw conic curves, including ones with an infinite gradient (such as the horizon line).
Also draws the horizon line, as well as the positive X direction.


TrentHouliston:
Looks good to myself!

PR #185, opened 2018-05-04
Enable decoding of jpeg images as video streams

TrentHouliston:


BrendanAnnable:
How do I test this?

src/client/components/vision/network.ts
```diff
@@ -9,10 +9,13 @@ import { RobotModel } from '../robot/model'
 import { CameraModel } from './camera/model'
 import { VisionRobotModel } from './model'
 import Image = message.input.Image
+import CompressedImage = message.output.CompressedImage
 
 export class VisionNetwork {
+
   constructor(private network: Network) {
     this.network.on(Image, this.onImage)
+    this.network.on(CompressedImage, this.onImage)
```
BrendanAnnable: Looks like the `image` param type on the `onImage` method should be `Image | CompressedImage`

BrendanAnnable: Or two separate methods that delegate.

PR #187, opened 2018-07-25
Change chart network lifetime to when in view

TrentHouliston:
The chart view can take up a lot of CPU power when it's not used.
This stops it from using that while you're not even looking at it.


TrentHouliston:
Looks good to myself!

PR #188, opened 2018-07-25
Improve performance of VisualMesh

TrentHouliston:
Change how the data is transferred in order to improve performance and update drawing.
Classifications still need to be drawn properly in a future PR


TrentHouliston:
Works, is much faster yay.
🙏 self five!

PR #189, opened 2018-07-25
Draw the Visual Mesh in the vision window

TrentHouliston:
Add drawing the VisualMesh in the vision window


TrentHouliston:
Merging, But at some point, someone should add an interface to show/hide various things

PR #192, opened 2018-08-12
Upgrade to Webpack 4 and TypeScript 2.9.2

BrendanAnnable:
😐


webpack.config.ts
```diff
@@ -130,29 +115,45 @@ export default [{
...
+          chunks: 'initial',
+          minChunks: 2,
+        },
+        vendor: {
+          test: /[\\/]node_modules[\\/]/,
+          chunks: 'all',
+          priority: -10,
+        },
+        proto: {
+          test: /src[\\/]shared[\\/]proto/,
```
BrendanAnnable: Keeps the proto definitions in their own chunk.

PR #193, opened 2018-08-12
TS: target ES2017

BrendanAnnable:
This means ES6 classes and things are no longer transpiled to the older ES5 syntax.


tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "sourceMap": true,
-    "target": "es5",
+    "target": "ES2017",
```
BrendanAnnable: All the above changes are re-ordering things so that dependencies defined prior to their use.

Apparently required to make this 1 line change.

PR #204, opened 2018-09-19
Enable strict property initialization

BrendanAnnable:
https://blog.mariusschulz.com/2018/05/20/typescript-2-7-strict-property-initialization

### DIFF
--- a/src/client/components/dashboard/dashboard_robot/model.ts
+++ b/src/client/components/dashboard/dashboard_robot/model.ts
@@ -71,9 +71,59 @@ export class DashboardRobotModel {
   @observable walkPathPlan: Vector2[]
   @observable walkCommand: Vector3
 
-  constructor(robot: RobotModel, opts: DashboardRobotModelOpts) {
+  constructor(robot: RobotModel, {
+    camera,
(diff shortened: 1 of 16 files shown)


src/client/components/dashboard/dashboard_robot/model.ts
```diff
@@ -71,9 +71,59 @@ export class DashboardRobotModel {
...
+    gameMode,
+    gamePhase,
+    penaltyReason,
+    playerId,
+    lastCameraImage,
+    lastSeenBall,
+    lastSeenGoal,
+    walkPathPlan,
+    walkCommand,
+  }: DashboardRobotModelOpts) {
```
BrendanAnnable: This class has a lot of dependencies! 😬 decomposing them would be a nice follow-up task.

tsconfig.json
```diff
@@ -10,7 +10,6 @@
     "declaration": false,
     "removeComments": true,
     "strict": true,
-    "strictPropertyInitialization": false,
```
BrendanAnnable: This is the main point of this PR

PR #205, opened 2018-09-19
Stop simulators when there are no clients

BrendanAnnable:


src/server/prod.ts
```diff
@@ -44,9 +50,19 @@ if (withVirtualRobots) {
...
+        // tslint:disable-next-line no-console
+        console.log('Client connected, starting simulator')
+        stopSimulators = virtualRobots.startSimulators()
+      } else if (stopSimulators) {
+        // tslint:disable-next-line no-console
+        console.log('No clients remain connected, stopping simulator')
+      }
+    },
+    { fireImmediately: true },
+  )
```
BrendanAnnable: - [ ] Remove duplication between `prod.ts` and `dev.ts`

src/server/dev.ts
```diff
@@ -73,7 +76,21 @@ function init() {
...
+          // tslint:disable-next-line no-console
+          console.log('Client connected, starting simulator')
+          stopSimulators = virtualRobots.startSimulators()
+        } else if (stopSimulators) {
+          // tslint:disable-next-line no-console
+          console.log('No clients remain connected, stopping simulator')
+        }
+      },
+      { fireImmediately: true },
+    )
```
BrendanAnnable: Uglier than needed I suspect. Perhaps `VirtualRobots` should instead expose a more flexible API that would be more suitable for reacting to an observable value.

Something like:

```ts
reaction(() => {
  () => now(frequency) && numConections > 0,
  enabled => enabled && virtualRobots.tick()
})
```

Will try this later.

PR #206, opened 2018-09-28
[WIP] Nugus model

ghost:
[Work in Progress]
Currently have the models working, but leg movement animations look off


BrendanAnnable:
Very cool! Just a heads up, I won't be able to review this until early next week.

Just quickly, the heroku deploy failed with this error:

```
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
```

Not sure how this diff could cause that though... is the model too big?

luxdosugi:
The model is too big, changing over to a smaller one now though

BrendanAnnable:
I'm not sure what is happening 🤣feel free to reassign me when it looks like a robot again! 😁

luxdosugi:
Haha, oops sorry about that, they look vaguely robot now though.

TrentHouliston:
Just FYI the alignment of the legs is still off.
In case you didn't know, you can hit space twice when in the 3d view and get a locked on view of the robot.
Also while these lower poly models are good for loading, the look like potato so we need to find some middle ground of better quality without having a giant file size.

src/client/components/localisation/nugus_robot/head/view_model.ts
```diff
@@ -29,7 +29,7 @@ export class HeadViewModel {
     const { geometry, materials } = this.neckGeometryAndMaterial
     const mesh = new Mesh(geometry, materials)
     mesh.position.set(0, 0.051, 0.08)
-    mesh.rotation.set(0, this.model.motors.headPan.angle, 0)
+    mesh.rotation.set(0, this.model.motors.headPan.angle, this.model.motors.headTilt.angle)
```
TrentHouliston: I very much doubt that you should be doing both of these rotations at once. Our servos don't have two degrees of freedom

luxdosugi: That's nice to know. If we call the y axis up, which axis can the head rotate on?

TrentHouliston: The axis of rotation will depend on the local frame of reference. The way to think about it is that the robot is made up of rigid bodies. Each of these is a part that is solidly attached to a single servo. This servo can only rotate around a single axis. All of the parts should just be rigid bodies in rigid bodies all centered on the torso. So for the head, there are three rigid bodies at play, the torso, which is always just itself, the neck, which is just a single servo with no printed parts, and the head. So you can rotate the neck, and you can rotate the head, but this is two single rotations.

TrentHouliston: So if you look in the Darwin model, all rotations are around the x or y axis. That’s because the rotation is local to the frame of reference. The rotation of the head may be x or y (or z) depending on the orientation of the actual part itself.

luxdosugi:
How does this look now?

TrentHouliston:
You're missing a whole bunch of the rigid bodies that are just servos (neck, hip).
And surely there is a compromise between the models looking horribly polygoned and megabytes of data?

luxdosugi:
I have just added the rigid bodies for the servos between moving joints,
Am I missing anything there?

TrentHouliston:
As far as I can tell you're still missing the servos in the neck and hips

TrentHouliston:
image

luxdosugi:
Touché, initially did not have the model for the servo, but I think I saw it somewhere.

luxdosugi:
I'm pretty sure that the legs are at least a little wrong, given that the feet pass through each other. I will look at the actual robot tomorrow or next week.

luxdosugi:
Probably has to do with the current model file size, but I have noticed that the ram steadily spikes up and eventually crashes the tab running NUsight2 localisation mode. On the right of the images there is RAM usage, the first image is just before crash, second  image is the crash and also shows normal RAM usage.
screenshot from 2018-12-07 15-41-42
screenshot from 2018-12-07 15-41-52

TrentHouliston:
That’ll be a problem with how you’re making your models. You’re probably loading a new one in constantly. The localisation tab never had its memory management fixed like in the vision window

luxdosugi:
Should fixing it be similar to the vision window?

TrentHouliston:
Yes, using create transformer along with a deallocator

luxdosugi:
Cool, I will look into that.

BrendanAnnable:
> The localisation tab never had its memory management fixed like in the vision window

You can blame me! 😄 I've been meaning to get around to cleaning up the localisation code.

TrentHouliston:
To be fair, you did all that before we even knew that was a thing

TrentHouliston:
Also, before this PR can be finished, you'll need to ensure that this model is modular along with the Darwin model so that we are able to switch between them easily. It's not as important for us to have the Darwin model, but it is important that we can do that when we get future robot models.

luxdosugi:
I guess have some choosing type menu, and separate code for the darwin and nugus models?

BrendanAnnable:
Is there nothing at the network layer to determine the type of robot?

LukeFarrawell:
If we are going to continue having it as a feature, the first person camera is still at the waist.

TrentHouliston:
No need, you can just stick the camera object itself in the head!

BrendanAnnable:
That does sound much easier

luxdosugi:
> Is there nothing at the network layer to determine the type of robot?

Probably, I have not looked at the network layer yet.

TrentHouliston:
There is nothing in the network layer that directly tells you robot type, but we do name our robots igus1 igus2 etc, so you can access that name to work it out

BrendanAnnable:
Superseded by #280

PR #207, opened 2018-10-06
More upgrades!

BrendanAnnable:


declarations/dom.d.ts
```diff
@@ -0,0 +1,7 @@
+interface Element {
+    requestPointerLock(): void;
+}
+
+interface Document {
+  pointerLockElement: Element;
+}
```
BrendanAnnable: Apparently this is the recommended fix.

https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#recommendations

PR #209, opened 2018-10-22
Add cross-env for correct env variables on Windows

JosephusPaye:
Setting environment variables like `TS_NODE_FILES=true` doesn't work in the native Windows shell (which `yarn`/`npm` use for running scripts).

This can be fixed by using [cross-env](https://github.com/kentcdodds/cross-env) and simply prepending `cross-env` like: `cross-env TS_NODE_FILES=true`.

I did a quick search for other areas we set env variables, but I may have missed something. Please add others as needed.


BrendanAnnable:
👍 LGTM.

Ideally we can remove the env var entirely, it was added when `ts-node` broke backwards compatibility and it was the easy fix.

I think you're supposed to run without it, somehow.

PR #213, opened 2018-12-26
Add DarwinRobot story

BrendanAnnable:
screen shot 2018-12-27 at 12 39 40 am

https://nusight-pr-213.herokuapp.com/storybook/


JosephusPaye:
The `ModelVisualizer` here would be very useful for the script tuner. Could we make it a standalone component for use elsewhere outside this story?

BrendanAnnable:
yeah I think we could probably re-use it 🙂

src/client/components/localisation/darwin_robot/stories/model_visualiser.tsx
```diff
@@ -0,0 +1,101 @@
...
+  private camera(canvas: Canvas) {
+    const camera = new PerspectiveCamera(75, canvas.width / canvas.height, 0.01, 100)
+    camera.position.set(0.4, 0.3, 0.3)
+    camera.up.set(0, 0, 1)
+    camera.lookAt(0, 0, 0)
+    return camera
+  }
+
+  private scene(children: Object3D[]) {
+    const scene = new Scene()
```
TrentHouliston: Isn't this a memory leak?
Each time making a new scene means the renderer doesn't invalidate objects in the old scene

src/client/components/localisation/darwin_robot/stories/three.tsx
```diff
@@ -0,0 +1,84 @@
...
+    this.copyToLocalScene(scene)
+    this.renderer!.setSize(this.canvas.width, this.canvas.height, false)
+    this.renderer!.render(this.scene, camera)
+  }
+
+  private copyToLocalScene(scene: Scene) {
+    // Internally, three.js keeps a mapping between scene objects and various resources.
+    // This means we need to pass the same scene object to the renderer every frame.
+    // To allow consumers of this component to create scene objects at will, we take on the burden of copying the
+    // children into our own permanent scene object.
```
TrentHouliston: You should put some comment like this where the new scene is maybe :P

src/client/components/localisation/darwin_robot/stories/model_visualiser.tsx
TrentHouliston: Well, it's actually solved by the thing I commented about below.
Each renderer caches objects by scene/camera combination. So if you change the scene it ends up with another copy. Dispose works, but if you're constantly caching the same object it'll make a new copy for every scene/camera combo. In those cases you won't dispose (because it hasn't changed) but because the scene changed it got another copy. Or at least I think that's how it worked

TrentHouliston: Although you should make sure that you're not creating new cameras each time too

PR #214, opened 2019-01-02
Add Vision Objects

BrendanAnnable:


src/client/components/vision/camera/view_model.ts
```diff
@@ -93,9 +97,40 @@ export class CameraViewModel {
     if (this.model.visualmesh) {
       scene.add(this.visualmesh(this.model.visualmesh))
     }
+    this.model.balls.forEach(ball => scene.add(this.ball(ball)))
```
TrentHouliston: These use Hcw in image, is this going to be a problem if an image hasn't loaded yet?
In theory you should be able to render balls on a blank image

src/client/components/vision/camera/view_model.ts
```diff
@@ -93,9 +97,40 @@ export class CameraViewModel {
...
+      lineWidth: 10,
+    })
+  })
+
+  private goal = createTransformer((m: Goal) => {
+    const goal = new Object3D()
+    const Hwc = new Matrix4().getInverse(toThreeMatrix4(m.Hcw))
+    const imageHcw = this.model.image ? toThreeMatrix4(this.model.image.Hcw) : new Matrix4()
+    const Hcc = imageHcw.multiply(Hwc)
+    const tl = toThreeVector3(m.frustum.tl).applyMatrix4(Hcc)
```
TrentHouliston: Should this maybe be a list/map?

src/client/components/vision/camera/view_model.ts
```diff
@@ -93,9 +97,40 @@ export class CameraViewModel {
...
+    const tr = toThreeVector3(m.frustum.tr).applyMatrix4(Hcc)
+    const bl = toThreeVector3(m.frustum.bl).applyMatrix4(Hcc)
+    const br = toThreeVector3(m.frustum.br).applyMatrix4(Hcc)
+    const colour = new Vector4(0.8, 0.8, 0, 0.5)
+    goal.add(this.makePlaneSegment({ start: tl, end: bl, colour, lineWidth: 10 }))
```
TrentHouliston: You could also make this a map function if there is something like pythons zip :P

src/client/components/vision/camera/view_model.ts
```diff
@@ -347,3 +382,16 @@ export class CameraViewModel {
     })
   }
 }
+
+function toThreeMatrix4(mat4: Matrix4Model): Matrix4 {
```
TrentHouliston: This is in more than one place now, should it be some utility function?

BrendanAnnable: when I see it used 3 times we can move it somewhere 🙂

PR #216, opened 2019-01-27
Add Three component

BrendanAnnable:
Adds a new `Three` component designed to be used anywhere we require rendering of three.js scenes.

Its responsibilities:
- Take a `createStage` which produces a computed stage (a scene and camera pair).
- Provide the `width` and `height` of the canvas via an observable canvas object to `createStage`, which it can then use to (reactively) generate the scene from.
- Automatically re-render itself when the computed stage changes (by using a reaction).
- Resize itself to the container it is placed within.
- Eventually manage the horrible `Scene` memory leak problem (separate PR)

This PR also adds a storybook for the component with example usage. A separate PR will address the various memory leaks (such as disposing scenes/materials/geometries).

Storybook: https://nusight-pr-216.herokuapp.com/storybook/?selectedKind=component.three&selectedStory=renders%20animated%20scene&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

### DIFF
--- a/src/client/components/three/stories/three.stories.tsx
+++ b/src/client/components/three/stories/three.stories.tsx
@@ -1,0 +1,120 @@
+import { storiesOf } from '@storybook/react'
+import { IComputedValue } from 'mobx'
+import { reaction } from 'mobx'
+import { observable } from 'mobx'
+import { action } from 'mobx'
+import { computed } from 'mobx'
+import { disposeOnUnmount } from 'mobx-react'
+import { now } from 'mobx-utils'
+import { Component } from 'react'
+import * as React from 'react'
+import { Light } from 'three'
+import { Object3D } from 'three'
+import { Material } from 'three'
+import { Geometry } from 'three'
+import { Camera } from 'three'
+import { PointLight } from 'three'
+import { MeshPhongMaterial } from 'three'
+import { Mesh } from 'three'
+import { BoxGeometry } from 'three'
+import { PerspectiveCamera } from 'three'
+import { Scene } from 'three'
+
+import { Vector2 } from '../../../math/vector2'
+import { Vector3 } from '../../../math/vector3'
+import { Stage } from '../three'
+import { Canvas } from '../three'
+import { Three } from '../three'
+
+type BoxModel = { color: string, size: number, position: Vector3, rotation: Vector3 }
+
+class BoxVisualiser extends Component<{ animate?: boolean }> {
+  @observable
+  private model = {
+    boxes: [
+      { color: 'red', size: 1, position: Vector3.of(), rotation: Vector3.of() },
+      { color: 'green', size: 1, position: Vector3.of(), rotation: Vector3.of() },
+      { color: 'blue', size: 1, position: Vector3.of(), rotation: Vector3.of() },
+    ],
+  }
+
+  componentDidMount() {
+    this.update(0)
+    this.props.animate && disposeOnUnmount(this, reaction(() => now('frame'), this.update))
+  }
+
+  @action.bound
+  update(now: number) {
+    const t = 2 * Math.PI * now / (20 * 1000)
+    const n = this.model.boxes.length
+    this.model.boxes.forEach((box, i) => {
+      const position = Vector2.fromPolar(1, i * 2 * Math.PI / n + t)
+      box.position.set(position.x, position.y, 0)
+      box.rotation.set(Math.cos(3 * t + i), Math.cos(5 * t + i), Math.cos(7 * t + i))
+    })
+  }
+
+  render() {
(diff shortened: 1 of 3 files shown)


TrentHouliston:
So this should be everywhere eventually?
Vision, Localisation, Script Tuner etc, anything that uses ThreeJS should use this component?

src/client/components/three/stories/three.stories.tsx
```diff
@@ -0,0 +1,120 @@
...
+    const geometry = computed(() => this.boxGeometry())
+    const box = (box: BoxModel) => {
+      const material = computed(() => this.boxMaterial(box))
+      return computed(() => this.box(box, geometry.get(), material.get()))
+    }
+    const camera = computed(() => this.camera(canvas))
+    const light = computed(() => this.light(camera.get()))
+    const boxes = computed(() => this.model.boxes.map(b => box(b).get()))
+    const scene = computed(() => this.scene([...boxes.get(), light.get()]))
+    return computed(() => ({ camera: camera.get(), scene: scene.get() }))
```
BrendanAnnable: It is important to note that `createStage` returns a `IComputedValue<Stage>`, not just a `Stage`.

Specifically, the stage is not actually created until the computed is used. e.g.

```ts
const stage = createStage(canvas); // No three.js objects have been created yet. All we have done is create the computation tree.
stage.get(); // This is the call which recursively creates the three.js Scene and Camera objects and their children.
```

Having the entire scene being computed like this allows the render to:

1. react to any observable changes, and;
2. only require the computed functions whose dependencies have changes to re-compute.

This gives us the declarative efficiency we need.

Perhaps `createStage` is a misleading name then, but `createComputedStage` seems quite long...

BrendanAnnable:
Yep I think they will all find it useful.

src/client/components/three/stories/three.stories.tsx
```diff
@@ -0,0 +1,120 @@
...
+    return camera
+  }
+
+  private scene(children: Object3D[]): Scene {
+    const scene = new Scene()
+    scene.add(...children)
+    return scene
+  }
+
+  private box(box: BoxModel, geometry: Geometry, material: Material): Mesh {
```
TrentHouliston: Not 100% sure I understand but aren't these supposed to be some flavour of computed?

src/client/components/three/three.tsx
```diff
@@ -0,0 +1,57 @@
...
+  render() {
+    return <ReactResizeDetector handleWidth handleHeight onResize={this.onResize}>
+      <div className={styles.container}>
+        <canvas ref={this.setRef} className={styles.canvas}/>
+      </div>
+    </ReactResizeDetector>
+  }
+
+  private renderStage({ scene, camera }: Stage) {
+    this.renderer!.setSize(this.canvas.width, this.canvas.height, false)
```
TrentHouliston: I'm pretty sure you don't need to set the canvas width/height here, I think this function sets those while handling DPI settings etc.
So you should setSize to the size you want in onResize (pretty sure)

src/client/components/three/stories/three.stories.tsx
```diff
@@ -0,0 +1,120 @@
...
+    ],
+  }
+
+  componentDidMount() {
+    this.update(0)
+    this.props.animate && disposeOnUnmount(this, reaction(() => now('frame'), this.update))
+  }
+
+  @action.bound
+  update(now: number) {
```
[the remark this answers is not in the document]
BrendanAnnable: Ah injecting the dependency lets me call it above with `0` for the non-animated story.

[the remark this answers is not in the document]
TrentHouliston: I understand!

src/client/components/three/stories/three.stories.tsx
```diff
@@ -0,0 +1,120 @@
...
+    boxes: [
+      { color: 'red', size: 1, position: Vector3.of(), rotation: Vector3.of() },
+      { color: 'green', size: 1, position: Vector3.of(), rotation: Vector3.of() },
+      { color: 'blue', size: 1, position: Vector3.of(), rotation: Vector3.of() },
+    ],
+  }
+
+  componentDidMount() {
+    this.update(0)
+    this.props.animate && disposeOnUnmount(this, reaction(() => now('frame'), this.update))
```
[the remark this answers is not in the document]
BrendanAnnable: This function is not rendering though, it is updating the model (think network layer).

The render function could definitely use it though. so I've add it there! 497a3b9

src/client/components/three/three.tsx
TrentHouliston: I see :P redundant code so mobx knows what's happening

PR #220, opened 2019-01-28
Add @disposableComputed

BrendanAnnable:
More memory leaks to disappear!

### DIFF
--- a/src/client/base/disposable_computed.ts
+++ b/src/client/base/disposable_computed.ts
@@ -39,7 +39,7 @@ const disposableComputedExpr = <T extends { dispose(): void; }>(fn: () => T): IC
 
 export const disposableComputed: DisposableComputed = function computed(arg1: any, arg2: any, arg3: any) {
   if (typeof arg2 === 'string') {
-    // Use as an annotation e.g. @disposableComputed
+    // Used as an annotation e.g. @disposableComputed
     return disposableComputedDecorator.apply(null, arguments as any)
   }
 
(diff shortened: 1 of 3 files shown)


src/client/base/disposable_computed.ts
```diff
@@ -2,12 +2,31 @@ import { IComputedValue } from 'mobx'
...
 import { computed } from 'mobx'
 
+function disposableComputedDecorator(target: any, propertyKey: string, descriptor: PropertyDescriptor) {
+  if (!descriptor) {
+    descriptor = Object.getOwnPropertyDescriptor(target, propertyKey)!
+  }
+  const originalMethod = descriptor.get!
+  // Lazily create and store the computed expression per instance.
+  // This allows us to bind each computed expression to the correct `this` context.
+  const expressionPerInstance = new WeakMap()
```
TrentHouliston: Pfft, someone should do this the proper way someday

src/client/components/three/stories/three.stories.tsx
```diff
@@ -139,7 +140,7 @@ class BoxViewModel {
     return mesh
   }
 
-  @computed
+  @disposableComputed
   private get material(): Material {
     return new MeshPhongMaterial({ color: this.model.color })
```
TrentHouliston: Does this still recompile the shader every time you change the colour?

BrendanAnnable: Yes 😭

TrentHouliston: And once again all of your wonderful abstractions must die in service of performance :P

BrendanAnnable: JUST DON'T CHANGE COLOURS OKAY EVERYONE 😭

BrendanAnnable: JUST DON'T DO ANYTHING USEFUL 😭

BrendanAnnable: I think we can solve this, but it might be tricky

PR #221, opened 2019-01-28
Reuse scene and camera objects

BrendanAnnable:
Fixes those pesky memory leaks.

### DIFF
--- a/src/client/components/three/three.tsx
+++ b/src/client/components/three/three.tsx
@@ -6,6 +6,7 @@ import { disposeOnUnmount } from 'mobx-react'
 import * as React from 'react'
 import { Component } from 'react'
 import ReactResizeDetector from 'react-resize-detector'
+import { Object3D } from 'three'
 import { WebGLRenderer } from 'three'
 import { Scene } from 'three'
 import { Camera } from 'three'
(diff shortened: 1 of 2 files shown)


src/client/components/three/three.tsx
```diff
@@ -54,3 +63,45 @@ export class Three extends Component<{ stage(canvas: Canvas): IComputedValue<Sta
...
+    this.copyToLocalScene(stage.scene)
+    this.copyToLocalCamera(stage.camera)
+    return { camera: this.camera, scene: this.scene }
+  }
+
+  private copyToLocalScene(scene: Scene) {
+    this.scene.copy(scene)
+    this.scene.children.length && this.scene.remove(...this.scene.children)
+    for (const child of scene.children) {
+      // Hack: Remove the parent so that three.js doesn't remove the children from the original scene object.
```
TrentHouliston: Do we care? Won't the ones using this be recreating the scene object each time anyway?

BrendanAnnable: Try removing this line and see what happens 😛 (all the stories go blank)

TrentHouliston: Seems important then! :P

BrendanAnnable: I agree that it should be regenerated 99% of the time, so I'll still look into why exactly this is needed.

...but I'll merge this for now 🙂

PR #223, opened 2019-01-30
Add LUT visualiser

BrendanAnnable:
screen shot 2019-01-30 at 6 07 38 pm

https://nusight-pr-223.herokuapp.com/storybook/?selectedKind=classifier.visualizer&selectedStory=renders%20animated&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

### DIFF
--- a/src/client/components/classifier/lut.ts
+++ b/src/client/components/classifier/lut.ts
@@ -19,11 +19,11 @@ export class Lut {
     this.size = size
   }
 
-  static of() {
+  static of(size: number) {
     return new Lut({
       atom: createAtom('Lut'),
-      data: new Uint8Array(2 ** (6 + 6 + 6)).fill(Classification.Unclassified),
-      size: { x: 6, y: 6, z: 6 },
+      data: new Uint8Array(2 ** (3 * size)).fill(Classification.Unclassified),
+      size: { x: size, y: size, z: size },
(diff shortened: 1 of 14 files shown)


src/client/components/classifier/lut.ts
```diff
@@ -19,11 +19,11 @@ export class Lut {
     this.size = size
   }
 
-  static of() {
+  static of(size: number) {
     return new Lut({
       atom: createAtom('Lut'),
-      data: new Uint8Array(2 ** (6 + 6 + 6)).fill(Classification.Unclassified),
```
TrentHouliston: Our LUTs these days aren't the same size in all dimensions.
We normally have like 6x5x5 luts etc rather than square ones.
Not sure if that matters for this of function though

TrentHouliston: And it seems this is gone already :P

TrentHouliston: Wait nope, It's there I just somehow commented on the wrong side of the diff?

BrendanAnnable: haha 😄 I'll update it to take 3 args ea52ff1583b09b776a99ff7379c6d2653155c43c

PR #226, opened 2019-02-03
Use Three component in localisation

BrendanAnnable:


src/client/components/three/computed.ts
```diff
@@ -0,0 +1,198 @@
...
+) => (expr: () => O): IComputedValue<T> => {
+  let instance: T
+  return computedFn(() => {
+    const opts = expr()
+    if (instance == null) {
+      instance = create(opts)
+    }
+    update(instance, opts)
+    return instance
+  }, { equals: () => false })
```
BrendanAnnable:
```ts
{ equals: () => false }
```

This is the magic I was missing. If I thought about this earlier I wouldn't have bothered with the reconciler 😄 

This allows consumers to recompute even if we return the same instance.

Should this ever return true? Probably not.

If this `equals` method is ever called, it means an observable dependency has changed, which means a 99% chance that we no longer equal. So there is little to zero reason to waste CPU cycles confirming that we don't. It's a win-win!

TrentHouliston: The only reason I would still thing equals true might be useful is if you have a dependency in your dependency that changed, and you yourself didn't change. In which case... does always returning equals true work too?

src/client/components/three/computed.ts
```diff
@@ -45,7 +45,6 @@ const computedInstance = <T, O>(
 const disposedComputed = <T extends { dispose(): void; }>(fn: () => T): IComputedValue<T> => {
   let latest: T | undefined
   const expr = computed(() => {
-    latest && latest.dispose()
```
TrentHouliston: Does this even do anything anymore without dispose?
Now it's basically the same as `computed(fn)` right?

BrendanAnnable: The `dispose` on `onBecomeUnobserved` is still used below.

BrendanAnnable: improved with c5af825e

TrentHouliston: At this point it's almost worth using atoms for this

BrendanAnnable: How would atoms help?

TrentHouliston: Atoms are designed to manage a resource and notify when things change, aren't they? Which is what we are doing here?

TrentHouliston: Fair enough

PR #232, opened 2019-02-07
Add support for sending messages from NUsight

JosephusPaye:
This adds support for sending messages from NUsight onto the network.

Added a new method `send(message: Message, target?: string): void` on `src/client/network/network.ts`, used by components in their local network file.

Informally tested using a simulator in #208.


src/client/network/network.ts
```diff
@@ -44,4 +50,16 @@ export class Network {
     }
     this.offNUClearMessages.clear()
   }
+
+  /**
+   * Send the given message on the network
+   */
+  send(message: Message, target?: string): void {
```
TrentHouliston: Why make a new Message type with a target as a 2nd parameter here over using NUClearNetSend?

src/server/nuclearnet/fake_nuclearnet_server.ts
```diff
@@ -3,6 +3,7 @@ import { NUClearNetSend } from 'nuclearnet.js'
 import * as XXH from 'xxhashjs'
 
 import { createSingletonFactory } from '../../shared/base/create_singleton_factory'
+import { hashType } from '../../shared/nuclearnet/nuclearnet_proxy_parser'
```
TrentHouliston: I don't know where this function should go, both of these places seem like a bad spot.
How do we not have a utility location yet :P

src/shared/nuclearnet/nuclearnet_proxy_parser.ts
```diff
@@ -35,8 +36,17 @@ export class Encoder {
           case 'unlisten':
             return callback([JSON.stringify(packet)])
 
+          case 'packet': {
+            const { id, data: [key, { target, type, payload, reliable }] } = packet
+            return callback([
+              JSON.stringify({ id, nsp, key, header: { target, type, reliable } }),
```
TrentHouliston: You're sending `type` as a string here, and adding a hashType below that isn't a part of a NUClearNetSend packet.
The decoding logic below will take that and add a .hash element to it afterwards that'll be ignored, so the hashType(type) here is doing nothing.
However, NUClearNetSend does accept a buffer or a string here, so maybe instead don't send the type in json and in the parser below add it as the type.

JosephusPaye: Updated the decoder to not add the unused hash.

src/server/nuclearnet/fake_nuclearnet_server.ts
JosephusPaye: Moved back :)

src/client/network/network.ts
JosephusPaye: Changed to accept a `NUClearNetSend`.

src/server/nuclearnet/fake_nuclearnet_server.ts
BrendanAnnable: > How do we not have a utility location yet :P

NEVER!!! 🙂

PR #235, opened 2019-02-20
Add Storybook decorators

JosephusPaye:
This adds a number of convenience decorators for stories

- Font (`conf/storybook/decorators/font.tsx`): Sets the font family used in NUsight.
- Fullscreen: (`src/client/components/storybook/fullscreen.tsx`): Reset body padding and set width and height to fill viewport
- Container: (`src/client/components/storybook/container.tsx`): Create a decorator with a max-width container with optional padding

Storybook: https://nusight-pr-235.herokuapp.com/storybook/


src/client/components/select/stories/select.stories.tsx
```diff
@@ -12,21 +13,16 @@ const actions = {
...
 
-const Container = ({ children }: { children: any }) => {
-  return <div style={{ maxWidth: '320px', fontFamily: 'Arial, sans-serif' }}>
-    {children}
-  </div>
-}
+const container = createContainerDecorator({ maxWidth: '350px' })
 
 storiesOf('components.select', module)
+  .addDecorator(container)
```
BrendanAnnable: I think this one is too specific to be generic, can we just inline the decorator here? e.g.

```ts
.addDecorator((story: RenderFunction) => {
  return <div style={{ maxWidth: 350 }}>{story()}</div>
})
```

src/client/components/storybook/fullscreen.tsx
```diff
@@ -0,0 +1,11 @@
+import { RenderFunction } from '@storybook/react'
+import * as React from 'react'
+
+import * as style from './fullscreen.css'
+
+export function fullscreen(story: RenderFunction) {
+  return <div className={style.fullscreen}>
+    <style>{`body { margin: 0; padding: 0; }`}</style>
```
BrendanAnnable: Clever! I'm sure there is a better way to achieve this haha, but this is probably fine for now.

Can you add a comment why this is needed?

src/client/components/select/stories/select.stories.tsx
JosephusPaye: Yeah, makes sense.

JosephusPaye: Done!

src/client/components/storybook/fullscreen.tsx
JosephusPaye: Done!

PR #234, opened 2019-02-18
Add Script Tuner

JosephusPaye:
This adds a visual script tuner to NUsight.

PR #238, opened 2019-02-20
Add single robot selector

JosephusPaye:
This adds a component to select a single robot, using the Select component.

Extracted from #234, for use in the Script Tuner.

**Stories:**

https://nusight-pr-238.herokuapp.com/storybook/?selectedKind=components.robot_selector_single&selectedStory=renders%20empty&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel


JosephusPaye:
> I'd like to kill the notion of view models for anything that isn't a complex tree (such as a three.js scene). For simple components like this we should be able to just use simple props like regular react.
> 
> This means the state should live outside the component and controlled by a parent store, which can be shared with the rest of the application

Makes sense 👍, will keep that in mind going forward.

PR #259, opened 2019-03-28
[WIP] Switch from JSON to glTF

codymlewis:
I just started writing this, does nothing different currently

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,7 @@ yarn-error.log*
 
 # NBS recordings
 recordings
+
+# Some vim junk
+tags
+Session.vim
--- a/.gitignore
+++ b/.gitignore
@@ -31,4 +31,5 @@ recordings
 
 # Some vim junk
 tags
+tags.*
 Session.vim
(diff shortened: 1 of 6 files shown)


codymlewis:
I seem to have broken a decent amount of localisation by updating three.

I also added an extra rule to the tslint , which ignores the no-submodule-imports for gltfloader, probably not optimal.

BrendanAnnable:
I believe the updated version of three.js has its own TypeScript type definitions, so we also need to remove `@types/three` from `devDependencies` in `package.json`.

Not sure how many backwards incompatible changes there has been since we updated though.

codymlewis:
I found that turning off all the models make the rest of localisation work, so changing over to gltf should work. once I figure it out

codymlewis:
So far I have managed to get something that compiles and runs properlyish (uncommitted), however the model do not appear, but do load (can tell from the console logs). The asynchronous loading is a bit troublesome and I have tried a few techniques to make it work, I will have to try some more later. The mobx fromResource has acted a bit strange for me, but I am probably using it wrong.

Since the new load function uses urls to the model resources, I am current using a folder `assets/models/` for the binary gltf files, is that an okay change?

TrentHouliston:
Closed in favor of #280

PR #260, opened 2019-06-14
Specifies version number for node.

Bidski:
Installing the newest version of node (v12) results in compilation errors for NUClearNet.js. v10 is known to work (travis also uses v10 for testing).

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 
 ## Dependencies
-- [Node](https://nodejs.org/en/download/)
+- [Node v10.16](https://nodejs.org/en/download/)
 - [Yarn](https://yarnpkg.com/en/docs/install)
 
 ## Installing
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -6,7 +6,7 @@
 
 
 ## Dependencies
-- [Node](https://nodejs.org/en/download/)
+- [Node v10.16](https://nodejs.org/en/download/)
```
BrendanAnnable:
```suggestion
- [Node](https://nodejs.org/en/download/)
```

I suggest removing this, otherwise we'll forget about it 😄

Bidski: I originally put it there because the README file implies that you need to download and install Node separately to yarn. In which case, it should be obvious in the README what version is required.

Does Node need to installed independently, or will yarn pull it in automatically?

BrendanAnnable: Fair enough, you are correct that node has to be installed first.

Let's leave it in for now then – @TrentHouliston can we fix `NUClearNet.js`?

BrendanAnnable:
https://github.com/Fastcode/NUClearNet.js/pull/5

PR #262, opened 2019-06-17
Fix raw-loader vertex and fragment imports

JosephusPaye:
#257 upgraded `raw-loader` to [v2.0.0](https://github.com/webpack-contrib/raw-loader/releases/tag/v2.0.0), which switched from CommonJS exports to ES Modules. Because we use * imports:

```ts
import * as bayerFragmentShader from './shaders/bayer.frag'
```

we get a JS object `{ default: '...' }` instead of the shader/fragment string. This was stringified and passed to ThreeJS which tried to load it as a WebGL program, causing an error.

This PR changes the imports to a default import and updates the shader import type declarations.

Also upgraded to `raw-loader` to [v3.0.0](https://github.com/webpack-contrib/raw-loader/compare/v2.0.0...v3.0.0), after testing with no errors. Only change in v3.0.0 is a bump of the minimum node version to v8.9.0.


src/client/components/localisation/skybox/view_model.ts
```diff
@@ -10,8 +10,8 @@ import { Vector3 } from 'three'
 import { PlaneBufferGeometry } from 'three'
 
 import { SkyboxModel } from './model'
-import * as SkyboxFrag from './skybox.frag'
-import * as SkyboxVert from './skybox.vert'
+import SkyboxFrag from './skybox.frag'
+import SkyboxVert from './skybox.vert'
```
BrendanAnnable: Ditto 🙂 ...and all of the ones below

package.json
```diff
@@ -134,6 +133,7 @@
     "nuclearnet.js": "^1.5.3",
     "pixi.js": "^4.8.7",
     "protobufjs": "^6.8.8",
+    "raw-loader": "^3.0.0",
```
BrendanAnnable: I believe this should remain in `devDependencies`, i.e. it is only used during building/development.

JosephusPaye:
Also relaxed the required node version by removing the upper bound. Had v11.15.0 which was being rejected by yarn.

package.json
```diff
@@ -5,7 +5,7 @@
   "private": true,
   "repository": "github:NUbots/NUsight2",
   "engines": {
-    "node": ">=10.0.0 <11.0.0",
+    "node": ">=10.0.0",
```
BrendanAnnable: This was just added 🤣 #260

TrentHouliston: Well apparently it’s just 12 that’s a problem!

BrendanAnnable: If it needs updating let's move it to another PR.

JosephusPaye: Reverted in https://github.com/NUbots/NUsight2/pull/262/commits/5cb557f38d2495c855fb05fded7bde091038eb82

PR #264, opened 2019-06-18
Add ability to toggle vision objects

JosephusPaye:
This PR adds a dropdown menu on each camera image that allows for toggling the objects drawn (image, horizon, visual mesh, balls, goals, etc).

For the dropdown menu, I've added a new component `SwitchesMenu` which renders a dropdown menu where every option has a switch that can be toggled.

Storybook: https://nusight-pr-264.herokuapp.com/storybook/?path=/story/components-switches-menu--renders-empty


src/client/components/switches_menu/menu_icon.svg
```diff
@@ -0,0 +1,2 @@
+
```
TrentHouliston: Look at this entire wasted byte!

JosephusPaye: LOL 😄

src/client/components/vision/camera/model.ts
```diff
@@ -49,6 +51,50 @@ export class CameraModel {
...
+    image: true,
+    compass: true,
+    horizon: true,
+    visualmesh: true,
+    balls: true,
+    goals: true,
+  }
+
+  @computed
+  get drawOptions() {
```
BrendanAnnable: Please move this computed prop to the component class, as this is not really part of the model.

i.e. models are just about data, they shouldn't have actions, or know specifics about the UI.

PR #267, opened 2019-06-20
Convert the visual mesh to use rays instead of pixel coordinates so they can track

TrentHouliston:


BrendanAnnable:
> No description provided.

Can you provide a description? 😛 I have no idea what this does

PR #269, opened 2019-06-21
Add message rate counters

JosephusPaye:
This adds a counters for each robot that keeps track of:

- total number of packets received, and the rate of packets per second
- total number of bytes received, and the rate of bytes per second

The stats are stored on the `RobotModel` and updated in [src/client/network/nusight_network.ts](https://github.com/NUbots/NUsight2/pull/269/files#diff-e5dc5d55ea6b0c138f1b33662052f043) when we get new messages.

For showing the stats, I've added a new Collapsible component with [accompanying stories](https://nusight-pr-269.herokuapp.com/storybook/?path=/story/components-collapsible--interactive).

Storybook: https://nusight-pr-269.herokuapp.com/storybook/?path=/story/components-robot-selector--renders-with-updating-stats
Simulated: https://nusight-pr-269.herokuapp.com/


src/client/components/robot/model.ts
```diff
@@ -1,4 +1,15 @@
+import throttle = require('lodash.throttle')
```
JosephusPaye: Lodash exports native ES modules. Using the default import style (`import * as throttle from 'lodash.throttle'`), TypeScript complains:

image


Switching to `import throttle = require('lodash.throttle')` compiles, but `tslint` complains:

```
ERROR: src/client/components/robot/model.ts:1:19 - require() style import is forbidden
```

We could enable `esModuleInterop` in `tsconfig.json` or make `tslint` ignore this line.

src/client/components/robot/model.ts
```diff
@@ -17,7 +32,44 @@ export class RobotModel {
...
     return new RobotModel(opts)
   }
 }
+
+type RateOpts = {
+  smoothing: number
+  unitTime: number
+}
+
+class Rate {
```
JosephusPaye: Put this in this file as I'm not sure where it should go.

JosephusPaye: Put this in this file as I'm not sure where it should go.

src/client/components/robot/model.ts
```diff
@@ -1,4 +1,15 @@
+import throttle = require('lodash.throttle')
 import { observable } from 'mobx'
+import { runInAction } from 'mobx'
+
+type RobotModelOpts = {
```
JosephusPaye: With the new `dataRates` property, the class RobotModel can no longer be used as an interface, so I added this for use in the constructor and `of()`

src/client/components/robot/model.ts
JosephusPaye: Fixed by switching to another package: `throttle-debounce`.

src/client/components/robot/model.ts
JosephusPaye: Moved to `./rate.ts`

src/client/components/robot/rate.ts
```diff
@@ -0,0 +1,58 @@
...
+  private currentUpdateCount: number = 0
+  private lastUpdateTime?: number
+
+  constructor({ smoothing, unitTime }: RateOpts, private clock: Clock) {
+    this.smoothing = smoothing
+    this.unitTime = unitTime
+  }
+
+  static of(opts: RateOpts) {
+    return new Rate(opts, getClock())
```
BrendanAnnable: This file is (and will always be) in `/client`, so it will always be a `BrowserSystemClock`. So I suggest we just avoid the indirection.

```suggestion
    return new Rate(opts, BrowserSystemClock)
```

src/client/components/robot/rate.ts
```diff
@@ -0,0 +1,58 @@
...
+        // Start the new culmulative sum for the next rate update
+        this.currentUpdateCount = newCount
+      }
+    }
+
+    this.lastUpdateTime = now
+  }
+
+  private commit = throttle(1000, (rate: number) => {
+    runInAction(() => {
```
BrendanAnnable: No need to use `runInAction` when the surrounding function can be statically marked as an action:

```ts
private commit = throttle(1000, action((rate: number) => this.rate = rate * this.unitTime))
```

src/client/components/robot/rate.ts
```diff
@@ -0,0 +1,58 @@
+import { observable } from 'mobx'
+import { runInAction } from 'mobx'
+import { throttle } from 'throttle-debounce'
+
+import { Clock, getClock } from '../../../shared/time/clock'
+
+type RateOpts = {
+  smoothing: number
+  unitTime: number
```
BrendanAnnable: What is a unit time?

src/client/components/robot_selector/robot_label/format-si.ts
```diff
@@ -0,0 +1,48 @@
...
+  '-6': 'µ',
+  '-9': 'n',
+  '-12': 'p',
+  '-15': 'f',
+  '-18': 'a',
+  '-21': 'z',
+  '-24': 'y',
+}
+
+export function formatSI(value: number) {
```
BrendanAnnable: This looks like something easily testable 🙂

src/client/components/robot/model.ts
```diff
@@ -1,14 +1,32 @@
...
+}
+
 export class RobotModel {
   @observable id: string
   @observable connected: boolean
   @observable enabled: boolean
   @observable name: string
   @observable address: string
   @observable port: number
+  @observable stats = {
```
BrendanAnnable: I'm not sure this belongs here... I think this data is too specific to live up at the top-model level.

This is network-specific, so it feels like something that should live in `/client/network/model.ts`, indicated by the fact that `/client/network/nusight_network.ts` will be updating the model anyway, so it should probably own the model data structure as well.

Can we move the model data there?

src/client/components/robot/rate.ts
JosephusPaye: All the internal rate calculcation stuff is in `ms`, but the user may want per second, per minute, etc. This unit time does the conversion. E.g. for `unitTime = 1000`, we get the rate in seconds. It's used when we commit the calculated rate to the public property.

BrendanAnnable: Ah right. Any particular reason we perform calculations in milliseconds? Could we prefer seconds? We have tried to use seconds everywhere throughout NUbots and NUsight. You'll note the `BrowserSystemClock` interface deals only in seconds.

src/client/components/robot/rate.ts
JosephusPaye: That's what I had initially. But that led to an issue I didn't know how to solve:

Since `Rate` is created indirectly when `RobotModel` is created, there's no opportunity to specify the type of clock to use.

But since it's a better idea to move the stats and rate stuff from `RobotModel`, we can use `BrowserSystemClock` here 👍.

src/client/components/robot/rate.ts
JosephusPaye: Nice 👍

src/client/components/robot/model.ts
JosephusPaye: So have a model property on `NUsightNetwork` that'll have a list of stats corresponding to the app-level `RobotModel`s?

If we do so, any ideas on how we could expose that to show in the UI? Like we do here: https://github.com/NUbots/NUsight2/blob/paye/robot-packet-rate/src/client/components/robot_selector/view.tsx#L45

src/client/components/robot_selector/robot_label/format-si.ts
JosephusPaye: :grinning: True. I'll add some tests...

src/client/components/robot/model.ts
JosephusPaye: I'm thinking something like (rough pseudocode):

In `network/nusight_network.ts`:

```ts
type RobotNetworkStats = {
  packets: number
  packetsPerSecond: Rate
  bytes: number,
  bytesPerSecond: Rate
}

class NUsightNetwork {
  stats: Map<string, RobotNetworkStats>

  constructor(..., private clock: Clock) {
    this.stats = new Map()
  }

  onNUClearJoin(cb: (peer: NUClearNetPeer) => void) {
    this.nuclearnetClient.onJoin((peer) => {
      cb(peer)

      const robot = findRobot(peer)

      this.stats.set(robot.id, {
        packets: 0,
        packetsPerSecond: Rate.of({ ... }, this.clock),
        // ...
      })
    })
  }

  onNUClearMessage<T>(messageType: MessageType<T>, cb: MessageCallback<T>) {
    //...
    return this.nuclearnetClient.on(messageTypeName, (packet: NUClearNetPacket) => {
      //...
      if (robotModel) {
        this.updateStats(this.stats.get(robotModel.id), {
          packets: 1,
          bytes: buffer.length,
        })
        cb(robotModel, message)
      }
    })
  }

  updateStats(robotModel, { bytes, packets }) {
    const stats = this.stats.get(robotModel.id)
    stats.packets += packets
    stats.packetsPerSecond.update(packets)
    stats.bytes += bytes
    stats.bytesPerSecond.update(bytes)
  }
}
```

In `robot_selector/robot_label/view_model.ts`:

```ts
class RobotLabelViewModel {
  //...
  @computed stats() {
    return this.network.stats.get(this.model.id) // has access to NUsightNetwork as `network` somehow?
  }
}
```

In `robot_selector/robot_label/view.tsx`:

```tsx
<div className={style.dataCell}>
  <div className={style.dataLabel}>Packets</div>
  <div className={style.dataValue}>{
    formatSI(viewModel.stats.packets)
  }</div>
</div>
```

Probably won't work, but is it like what you had in mind?

src/client/components/robot/rate.ts
JosephusPaye: No specific reason, was incidentally because `(Date|performance).now()` returns milliseconds.

Will change to seconds to match the `Clock` interface.

JosephusPaye: Also probably won't need /minute, /hour, etc. so I'll remove the `unitTime` option.

src/client/components/robot_selector/robot_label/format-si.ts
JosephusPaye: Added in https://github.com/NUbots/NUsight2/pull/269/commits/f68f92f91a4c839951594c6c47cfbfcb240f09a1

src/client/components/robot/rate.ts
JosephusPaye: Updated in https://github.com/NUbots/NUsight2/pull/269/commits/d45eebc07e27f8ee6ec63ec12ad609812f2ae61b

src/client/components/robot/rate.ts
JosephusPaye: Updated in https://github.com/NUbots/NUsight2/pull/269/commits/d45eebc07e27f8ee6ec63ec12ad609812f2ae61b

src/client/components/robot/rate.ts
JosephusPaye: Updated in https://github.com/NUbots/NUsight2/pull/269/commits/d45eebc07e27f8ee6ec63ec12ad609812f2ae61b

src/client/components/robot/model.ts
BrendanAnnable: I recommend keeping the model separate to the `NUsightNetwork` class.

We already do this for all the other model types. So I would expect something like this:

e.g.
```ts
class RobotNetworkStatsModel {
  @observable.ref packets: number
  ...

  static of = memoize((robotModel: RobotModel) => {
    return new RobotNetworkStatsModel(robotModel)
  })
}
```

```ts
class NUsightNetwork {
  @action
  onNUClearMessage() {
    const stats = RobotNetworkStatsModel.of(robotModel)
    stats.packets++
  }
}
```

```ts
class MyComponent extends Component<{ robotModel: RobotModel }> {
  @computed
  private get stats() {
    // Note: It would probably be more correct to pass this dependency in as a prop, but that might be difficult with the current architecture of the robot selector.
    return RobotNetworkStatsModel.of(this.props.robotModel)
  }

  render() {
    return <div>{this.stats.packets}</div>
  }
}
```

JosephusPaye: This is much simpler, thanks!

JosephusPaye: Implemented in https://github.com/NUbots/NUsight2/pull/269/commits/c9e872156322db2df91be7b947247c2ebbb24f94

src/client/components/robot_selector/robot_label/view.tsx
```diff
@@ -0,0 +1,76 @@
...
+  selectRobot(robot: RobotModel): void
+}
+
+export const RobotLabel = observer((props: RobotLabelProps) => {
+  const { robot, selectRobot } = props
+  const viewModel = RobotLabelViewModel.of(props.robot)
+
+  const toggleRobot = (robot: RobotModel) => () => selectRobot(robot)
+  const toggleStats = action((event: React.MouseEvent<HTMLButtonElement>) => {
+    event.stopPropagation()
```
BrendanAnnable: Can we avoid this? 😄

src/client/components/robot_selector/robot_label/view.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { RobotLabelViewModel } from './view_model'
+
+export type RobotLabelProps = {
+  robot: RobotModel
+  selectRobot(robot: RobotModel): void
+}
+
+export const RobotLabel = observer((props: RobotLabelProps) => {
+  const { robot, selectRobot } = props
+  const viewModel = RobotLabelViewModel.of(props.robot)
```
BrendanAnnable: I think I'm aiming to remove the concept of view models altogether, I really only want to use them to generate three.js scenes — maybe I should rename them to something else.

Anyway I think we can just use `RobotNetworkStatsModel.of(this.robotModel)` here 🙂

src/client/components/robot_selector/robot_label/view_model.tsx
```diff
@@ -0,0 +1,22 @@
+import { observable } from 'mobx'
+import { computed } from 'mobx'
+import { createTransformer } from 'mobx-utils'
+
+import { RobotNetworkStatsModel } from '../../../network/robot_network_stats_model'
+import { RobotModel } from '../../robot/model'
+
+export class RobotLabelViewModel {
+  @observable statsOpen = false
```
BrendanAnnable: View models shouldn't have any state, this should go on a plain model somewhere..

PR #272, opened 2019-06-22
Fix localisation noclip controls

BrendanAnnable:


src/client/components/localisation/controller.ts
```diff
@@ -25,6 +21,7 @@ export class LocalisationController {
   onAnimationFrame(model: LocalisationModel, time: number) {
     model.time.time = time / 1000
     this.updatePosition(model)
+    model.time.lastPhysicsUpdate = time / 1000
```
BrendanAnnable: Apparently I just forgot to update the timestamp

src/client/components/localisation/model.ts
```diff
@@ -11,22 +11,22 @@ import { SkyboxModel } from './skybox/model'
 
 export class TimeModel {
   @observable time: number // seconds
-  @observable lastRenderTime: number // seconds
+  @observable lastPhysicsUpdate: number // seconds
```
BrendanAnnable: I have decoupled the physics loop from rendering loop — so this makes more sense.

src/client/components/localisation/controller.ts
```diff
@@ -221,8 +218,8 @@ export class LocalisationController {
     // This camera position hack will not work with orientation/head movement.
     // TODO (Annable): Sync camera position/rotation properly using kinematic chain.
     model.camera.position.set(target.rWTt.x, target.rWTt.y, target.rWTt.z + 0.15)
-    const Rwt = new Quaternion(target.Rwt.x, target.Rwt.y, target.Rwt.z, target.Rwt.w)
-    const heading = new Euler().setFromQuaternion(Rwt).z
+    const Rwt = new THREE.Quaternion(target.Rwt.x, target.Rwt.y, target.Rwt.z, target.Rwt.w)
+    const heading = new THREE.Euler().setFromQuaternion(Rwt).z
```
BrendanAnnable: Preferring to prefix `THREE` types when intermixed with our native types.

TrentHouliston: There are other places that we could do this too if this is the way we want to go

PR #275, opened 2019-06-29
Upgrade three.js to 0.106.2

BrendanAnnable:


package.json
```diff
@@ -71,7 +71,6 @@
     "@types/socket.io-client": "^1.4.29",
     "@types/storybook__addon-actions": "^3.4.3",
     "@types/storybook__react": "^4.0.2",
-    "@types/three": "^0.91.0",
```
BrendanAnnable: Types are now built in!

src/client/components/localisation/utils.ts
```diff
@@ -1,9 +1,9 @@
 import { createTransformer } from 'mobx-utils'
 import { MeshLambertMaterial } from 'three'
 import { Color } from 'three'
-import { JSONLoader } from 'three'
 import { Geometry } from 'three'
 import { Material } from 'three'
+import { LegacyJSONLoader } from 'three/examples/jsm/loaders/deprecated/LegacyJSONLoader'
```
BrendanAnnable: Yay we can import from three.js examples now 😄

src/client/image_decoder/image_decoder.ts
```diff
@@ -153,7 +153,8 @@ export class ImageDecoder {
       scene.remove(...this.bayerDecoder!.scene.children)
       scene.add(mesh)
 
-      this.renderer.render(scene, this.bayerDecoder!.camera, target)
+      this.renderer.setRenderTarget(target)
+      this.renderer.render(scene, this.bayerDecoder!.camera)
```
BrendanAnnable: The render method no longer takes a render target

tslint.json
```diff
@@ -53,6 +53,7 @@
     "no-reference-import": true,
     "no-require-imports": true,
     "no-shadowed-variable": false,
+    "no-submodule-imports": [true, "three/examples/jsm"],
```
BrendanAnnable: Submodule imports are banned by default, so this adds threejs examples to the whitelist of exceptions.

src/client/components/three/builders.ts
```diff
@@ -152,8 +152,7 @@ export const shaderMaterial = createUpdatableComputed(
 )
 
 type TypedArray
-  = ArrayBuffer
```
BrendanAnnable: `ArrayBuffer` is no longer allowed

src/client/components/localisation/utils.ts
```diff
@@ -33,7 +33,7 @@ const coloredMaterial = (material: Material, color?: string) => {
 }
 
 const parseConfig = createTransformer((config: any) => {
-  const { geometry, materials } = new JSONLoader().parse(config)
+  const { geometry, materials } = new LegacyJSONLoader().parse(config, '/')
```
BrendanAnnable: The types require a `path` to be given, but we have no external resources to load anyway. So it is not actually used.

package.json
TrentHouliston: Hurray!

tslint.json
TrentHouliston: Hopefully, we won't need to use examples anymore

src/client/components/vision/camera/view.tsx
```diff
@@ -72,6 +72,7 @@ export class CameraView extends Component<{ viewModel: CameraViewModel }> {
     const { viewModel } = this.props
     const renderer = viewModel.renderer(viewModel.canvas)
     if (renderer) {
+      renderer.setRenderTarget(null)
```
BrendanAnnable: I'm actually surprised it worked without this, but I believe this is the correct thing to do.

(and all goes away with my refactor)

PR #278, opened 2019-06-29
Update ball rendering to suit new ball detector in NUbots

Bidski:
Screenshot from 2019-06-29 21-17-06


src/client/components/vision/camera/model.ts
```diff
@@ -34,7 +35,9 @@ export interface Ball {
   readonly cone: {
     readonly axis: Vector3
     readonly gradient: number
```
TrentHouliston: This is no longer used, and radius is doing exactly what gradient did before...

PR #281, opened 2019-07-03
Add extra charts and remove unneeded non null assertion operators

TrentHouliston:


Bidski:
[NUbots PR #320](https://github.com/NUbots/NUbots/pull/320) will change the foot down interpretation. Once that PR is merged this PR will need to be updated accordingly.

PR #283, opened 2019-07-04
Add file tree component

JosephusPaye:
Continuing the break up of #234 (Script Tuner), this PR adds a file tree component for browsing and selecting a file from a list of files.

Storybook: https://nusight-pr-283.herokuapp.com/storybook/?path=/story/components-file-tree--interactive


TrentHouliston (approved):
I can't speak to if there is a better way to do this but the code seems sane and it works.
If you want to know if there is a better way to code parts you'll have to wait for @BrendanAnnable

PR #285, opened 2019-07-06
Add render target support to three builders

BrendanAnnable:
[Storybook](https://nusight-pr-285.herokuapp.com/storybook/?path=/story/component-three--renders-animated-scene-with-render-targets)

)Screen Shot 2019-07-06 at 4 18 43 pm


BrendanAnnable:
Monica said LGTM

PR #288, opened 2019-07-12
Optimise TypeScript compile options

BrendanAnnable:
I've been meaning to do this for a while.

The main crux of this PR is removing `TS_NODE_FILES=true` from the `ts-node` commands. Why is this beneficial (other than being recommended)?

Before:
- Running `yarn dev` (and `yarn build`) parsed and typechecked the whole project, including tests and any other files not imported by `dev.ts`, slowing down startup and also blocking development if tests are broken from your changes.

After:
- Running `yarn dev` (and `yarn build`) only parses and typechecks what `dev.ts` itself needs to run (by transitively following the imports), resulting in a faster startup time and better development cycle. Note that `yarn test` will typecheck tests, and so will `yarn tscheck`.

This meant moving the files in `/declarations` to the "standard" directory structure, which is `typings/module_name/index.d.ts` — so that is why there is a bunch of file moves in this PR.

I also deleted our own `xxhashjs.d.ts` in favour of the official types (I'll probably do more of these in another PR).

Ref: https://github.com/TypeStrong/ts-node#help-my-types-are-missing


JosephusPaye:
This is very nice, thanks! 👍

PR #289, opened 2019-07-12
Add ts-loader to replace awesome-typescript-loader

BrendanAnnable:
Happens to get rid of babel in the normal webpack workflow — unfortunately storybook still need it.

Before:

`yarn build  78.97s user 4.02s system 202% cpu 40.921 total`

After:

`yarn build  41.30s user 2.46s system 154% cpu 28.250 total`


webpack.config.ts
```diff
@@ -43,16 +43,13 @@ const config: webpack.Configuration = {
         exclude: [
           path.resolve(__dirname, 'node_modules'),
         ],
-        use: [{
-          loader: 'awesome-typescript-loader',
+        use: {
+          loader: 'ts-loader',
           options: {
-            useBabel: true,
```
BrendanAnnable: Yay, faster builds

webpack.config.ts
```diff
@@ -43,16 +43,13 @@ const config: webpack.Configuration = {
...
-            useCache: true,
-            forceIsolatedModules: true,
+            onlyCompileBundledFiles: true,
```
BrendanAnnable: Yay, even faster builds!

webpack.config.ts
```diff
@@ -103,19 +100,16 @@ const config: webpack.Configuration = {
       {
         test: /\.svg$/,
         exclude: /\.file.svg$/,
-        use: [
-          'babel-loader',
```
BrendanAnnable: Yay, even more faster builds!

PR #293, opened 2019-08-23
Upgrade SVG loader

BrendanAnnable:


webpack.config.ts
```diff
@@ -99,10 +99,10 @@ const config: webpack.Configuration = {
         exclude: /\.file.svg$/,
         use: {
           loader: 'react-svg-loader',
-          query: {
+          options: {
             svgo: {
               // svgo options
-              plugins: [{ removeTitle: true }],
+              plugins: [{ removeTitle: true }, { removeViewBox: false }],
```
BrendanAnnable: An "optimisation" that removes the svg `viewBox` attribute is now enabled by default after this version upgrade.

Unfortunately this optimisation breaks things, like the robot icon.

So I disabled it.

PR #296, opened 2019-08-24
Upgrade Socket IO

BrendanAnnable:


src/shared/nuclearnet/nuclearnet_proxy_parser.ts
```diff
@@ -22,6 +22,7 @@ export class Encoder {
   encode(packet: Packet, callback: (wire: any[]) => void) {
     switch (packet.type) {
       case TYPES.EVENT:
+      case TYPES.BINARY_EVENT:
```
BrendanAnnable: For reasons I don't understand, packets that were previously being marked as `EVENT` are now marked as `BINARY_EVENT`

TrentHouliston: Interesting, looks like the typings for these packets has become more advanced. Previously in order to work out the event was a "binary" event you had to look at the type of the message. I'm not 100% sure but if this is still broken, it could be that the binary events should be going to the default of this switch.

BrendanAnnable: I can't see anything broken from this change?

PR #297, opened 2019-09-06
Make math data structures immutable

BrendanAnnable:


JosephusPaye:
@BrendanAnnable just curious, what's the advantage of making these immutable? 🧐

BrendanAnnable:
To be fair, I'm still weighing the trade-off of it. Immutable data structures come with a number of advantages, to quote [immutable js](https://github.com/immutable-js/immutable-js#immutable-collections-for-javascript):

> Immutable data cannot be changed once created, leading to much simpler application development, no defensive copying, and enabling advanced memoization and change detection techniques with simple logic. Persistent data presents a mutative API which does not update the data in-place, but instead always yields new updated data

You can see that within this PR changeset, a number of defensive `clone()` calls go away. The set/clone/copy methods all get deleted since they are no longer needed. In practice it makes functions more [pure](https://en.wikipedia.org/wiki/Pure_function), which makes logic easily testable (typically great to do with maths operations). It also makes debugging quite nice, as you can run pure functions multiple times (e.g. in console) with various arguments with full confidence that your application state won't be changing.

But tbh the main motivation was actually the `Vector3::normalize()` method throwing in reactions. I can see `Vector2::normalize()` just inlined the length calculation to avoid this issue, and it seems that patch didn't make it to `Vector3`. However, I think it would be simpler just to remove `@computed` entirely from `/math` to solve that issue directly.

There are only three advantages I can think of by keeping these data structures mutable:

1. Fine-grained reactions, and thus performance gains — e.g. if a reaction only uses `vector.z`, it won't re-run if `vector.x` is changed. I think this is actually a _very_ weak argument, because the chances of having any discernible performance improvements are extremely low. I don't know of anything in NUsight that would currently benefit in any meaningful way from this.
2. Lower memory usage — since you can change `vector.x` without creating an entirely new vector. In practice I think this makes very little difference to NUsight — everything is coming over the network as brand new deserialized data, we are allocating brand new data all the time — I doubt we would see much benefit, but this is perhaps more debatable.
3. More self-contained — I can pass a math structure directly into a controller for it to update (e.g. `vector.set(x, y, z);`), but if its immutable, I have to pass in the _containing_ model for the controller to update (e.g. `model.vector = new Vector(x, y, z);`. I think this might actually be a good thing, as with the latter, you can easily trace which model is actually getting updated, rather than a random math data structure, which you have to manually go trace back to the model its attached to.

On balance I think I _prefer_ the immutable versions, but I no longer need it if I remove the `@computed` annotations.

JosephusPaye:
Well explained and very insightful! 👍 

It makes sense to go with the immutable versions, since as you've mentioned the memory and minor performance hits are less of a concern in NUsight.

Being able to use the math data structures without having to make clones all the time or worrying about accidentally mutating the wrong thing is a big win for ergonomics.

PR #298, opened 2019-11-19
Update image decoder to handle permuted bayer image formats

Bidski:
@TrentHouliston tried to be clever and came up with a new method of compressing bayer image formats into a JPEG container. This PR updates the image decoder to handle these formats.


src/client/image_decoder/image_decoder.ts
```diff
@@ -110,80 +133,144 @@ export class ImageDecoder {
...
+      case fourcc('PGR8'): // Polarized Colour Bayer
+      case fourcc('PGB8'): // Polarized Colour Bayer
+      case fourcc('PJBG'): // Permuted Polarized Colour
+      case fourcc('PJRG'): // Permuted Polarized Colour
+      case fourcc('PJGR'): // Permuted Polarized Colour
+      case fourcc('PJGB'): // Permuted Polarized Colour
+        return 4 // Four values per width/height
 
+      default:
+        return 0
```
TrentHouliston: Maybe this should default to 1

src/client/image_decoder/image_decoder.ts
```diff
@@ -33,33 +33,37 @@ export interface Image {
...
 export class ImageDecoder {
-
   private atom: IAtom
 
   // These two functions are used to cache/destroy textures as they are created
   private cache: {
-    get(): Texture
-    destroy(): void
+    get(): Texture;
+    destroy(): void;
```
TrentHouliston: Semicolons?! I thought we didn't have those in NUsight2. Doesn't tslint tell you to get rid of them?

JosephusPaye: As statement terminators yes, but probably not in type definitions like here.

PR #302, opened 2020-01-15
Use BuildKite for NUsight2 builds

TrentHouliston:


BrendanAnnable:
Is buildkite better somehow? 😮

TrentHouliston:
It might be! It at the very least isn’t running on the terrible vms that travis uses. But it’s impossible to know without making a PR so it can try do things. We use it for the NUbots repo and builds went from 40 minutes to 4

TrentHouliston:
It's currently faster than Travis was, however, it's not doing the same caching of node_modules etc. So it could be made even faster in theory if someone wants to look into that in the future. But for now faster is faster.

BrendanAnnable:
Can we keep the slack integration?

JosephusPaye:
Can do: https://buildkite.com/docs/integrations/slack

TrentHouliston:
Slack integration has been done

PR #304, opened 2020-02-11
Add server-side compilation build

BrendanAnnable:
This makes `yarn build` compile both the client and server code.

`yarn prod` should be faster, as it now runs a pre-compiled JS file.

This also means deploy targets (e.g. heroku) no longer need dev dependencies to run the deploy.


BrendanAnnable:
I've broken the dedicated `yarn virtual-robots` command since it's not going through webpack.

...but since I'm pretty sure I'm the only one who uses that, and this PR is unblocking a bunch of things, I'm going to merge anyway and fix later.

PR #308, opened 2020-02-14
Add math three.js helpers

BrendanAnnable:
Since we do it everywhere, this PR at least makes it relatively easy to convert between the types.

Added some `toString` helpers as well.

(p.s. if only we had typeclasses)


BrendanAnnable:
> But did you also want to convert usages to use these where needed?

I do! But separately 🙂

PR #310, opened 2020-02-15
Add image view

BrendanAnnable:
A reusable component and view model for rendering images in all supported formats. This will replace `ImageDecoder`.

https://nusight-annable-image-v-advd2c.herokuapp.com/storybook/?path=/story/components-vision2-image-view--jpeg


src/client/components/vision2/image_view/stories/image_view.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+  .add('JPRG', () => {
+    const model = createModel()
+    loadImageElement(jprgUrl, ImageFormat.JPRG).then(action((image: Image) => model.image = image))
+    return <Observer>{() => <>
+      {model.image ? <ImageView image={model.image}/> : null}
+    </>}</Observer>
+  })
+  .add('RGGB', () => {
+    const model = createModel()
+    loadImageData(rggbUrl, 1280, 1024, ImageFormat.RGGB).then(action((image: Image) => model.image = image))
```
TrentHouliston: This is a pretty big file to be committing to the repo, maybe instead we should generate a bayer pattern image on the fly from a regular image. So converting
`RGB` `RGB` `RGB` `RGB`
`RGB` `RGB` `RGB` `RGB`
to
`R` `G` `R` `G`
`G` `B` `G` `B`

for an `RGGB` bayer pattern

PR #312, opened 2020-02-16
Add green horizon view model and storybook

BrendanAnnable:
Just look at this wonderfully realistic green horizon

Screen Shot 2020-02-16 at 3 14 55 pm

https://nusight-annable-green-h-q0sxs6.herokuapp.com/storybook/?path=/story/components-vision2-camera-greenhorizon--renders-animated


src/client/components/vision2/camera/greenhorizon.ts
```diff
@@ -0,0 +1,63 @@
...
+    end: this.rays[index + 1],
+    color: new Vector4(0, 0.8, 0, 0.8),
+    lineWidth: 10,
+  }))
+
+  @computed
+  private get rays() {
+    const { horizon, Hcw: oldHcw } = this.model()
+    const newHcw = this.imageHcw()
+    const oldHwc = Matrix4.fromThree(new THREE.Matrix4().getInverse(oldHcw.toThree()))
```
TrentHouliston: I think old and new aren't the best names here. It's more like, `imageHcw` or `projectionHcw` rather than new and `measurementHcw` or `rayHcw`. Or some other synonym for those. There's no reason that the GreenHorizon Hcw isn't newer than the image one in terms of time. Also i'm not 100% sure on how to fix it but sticking things on the Hcw etc variables always looks weird

src/client/components/vision2/camera/greenhorizon.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  private readonly segment = coneSegment((index: number) => ({
+    start: this.rays[index],
+    end: this.rays[index + 1],
+    color: new Vector4(0, 0.8, 0, 0.8),
+    lineWidth: 10,
+  }))
+
+  @computed
+  private get rays() {
```
TrentHouliston: This function probably needs some comments as to what it is doing as it's not immediately clear if you didn't do the math yourself :P

src/client/components/vision2/camera/model.ts
```diff
@@ -0,0 +1,52 @@
...
+  ) {
+    this.image = image
+    this.Hcw = Hcw
+    this.lens = lens
+  }
+}
+
+export interface Lens {
+  projection: Projection,
+  normalizedFocalLength: number,
```
TrentHouliston: I think it's still just focalLength. Normalised just describes what units it's in. And since it's normalised all through this codebase adding that is a little redundant.

src/client/components/vision2/camera/model.ts
```diff
@@ -0,0 +1,52 @@
...
+  centre?: Vector2,
+  distortionCoeffecients?: Vector2,
```
TrentHouliston:
```suggestion
  distortionCoefficients?: Vector2,
```

src/client/components/vision2/camera/shaders/world_line.frag
```diff
@@ -0,0 +1,280 @@
...
+uniform vec4 color;
+uniform float lineWidth;
+
+varying vec2 vUv;
+
+#define RECTILINEAR_PROJECTION 1
+#define EQUIDISTANT_PROJECTION 2
+#define EQUISOLID_PROJECTION 3
+
+// TODO(trent) these should be moved into a separate GLSL file once there is a decent #include system
```
TrentHouliston: Does one of these exist yet? :P

src/client/components/vision2/camera/shaders/world_line.frag
```diff
@@ -0,0 +1,280 @@
...
+ *
+ * @return an distortion radius
+ */
+float distort(float r, vec2 k) {
+  // Uses the math from the paper
+  // An Exact Formula for Calculating Inverse Radial Lens Distortions
+  // https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4934233/pdf/sensors-16-00807.pdf
+  // These terms have been stripped back to only include k1 and k2 and only uses the first 4 terms
+  // if more are needed in the future go and get them from the original paper
+  // TODO if performance ever becomes an issue, this can be precomputed for the same k values
```
TrentHouliston: You might want to compute b1-4 on the javascript side and upload it as a uniform.

src/client/components/vision2/camera/world_line.ts
```diff
@@ -0,0 +1,136 @@
...
+import { rawShader } from '../../three/builders'
+import { shaderMaterial } from '../../three/builders'
+import { createUpdatableComputed } from '../../three/create_updatable_computed'
+import { Canvas } from '../../three/three'
+
+import { Lens } from './model'
+import fragmentShader from './shaders/world_line.frag'
+import vertexShader from './shaders/world_line.vert'
+
+/** If start === end this will draw the entire cone. */
```
TrentHouliston: This comment seems a little out of place? Should it be somewhere else?

src/client/components/vision2/camera/world_line.ts
```diff
@@ -0,0 +1,136 @@
...
+    private readonly canvas: Canvas,
+    private readonly lens: () => Lens,
+  ) {
+  }
+
+  static of(canvas: Canvas, lens: () => Lens) {
+    return new WorldLine(canvas, lens)
+  }
+
+  /** Draws a segment of a plane projected to infinity in world space. */
```
TrentHouliston: Why does plane segment get a comment, but not the others?

src/client/components/vision2/camera/stories/greenhorizon.stories.tsx
```diff
@@ -0,0 +1,105 @@
...
+import { orthographicCamera } from '../../../three/builders'
+import { stage } from '../../../three/builders'
+import { Canvas } from '../../../three/three'
+import { Three } from '../../../three/three'
+import { GreenHorizonViewModel } from '../greenhorizon'
+import { Lens } from '../model'
+import { Projection } from '../model'
+import { GreenHorizon } from '../model'
+
+storiesOf('components.vision2.camera.greenhorizon', module)
```
TrentHouliston: Were you going to have the story of the actual image as well?

BrendanAnnable: #313 😄

src/client/components/vision2/camera/shaders/world_line.frag
BrendanAnnable: Perhaps! https://threejs.org/docs/#api/en/renderers/shaders/ShaderChunk

BrendanAnnable: Never mind, that is not what I thought it was 😞

src/client/components/vision2/camera/world_line.ts
```diff
@@ -0,0 +1,137 @@
...
+    const start = Vector3.fromThree(
+      (!axis.x && !axis.y)
+        ? new THREE.Vector3(0, 1, 0)
+        : new THREE.Vector3(-axis.y, axis.x, 0).normalize(),
+    )
+    return this.coneSegment({ axis, start, end: start, color, lineWidth })
+  }
+
+  /** Draws a segment of a plane projected to infinity in world space. */
+  planeSegment(segment: ConeSegment) {
```
TrentHouliston: I think it's incorrect to make this take a ConeSegment.
In a ConeSegment the axis variable should not be optional but here it is.

src/client/components/vision2/camera/world_line.ts
```diff
@@ -0,0 +1,137 @@
...
+    }
+  })
+
+  private static readonly shader = rawShader(() => ({ vertexShader, fragmentShader }))
+
+  private static readonly geometry = planeGeometry(() => ({ width: 2, height: 2 }))
+
+}
+
+export const coneSegment = createUpdatableComputed(
```
TrentHouliston: For the actual ConeSegment the axis is not optional.
Also, I'm not sure if I understood the updatable computed but should the arguments to this function be split into individual parameters? Or is it ok to have them in an object.

src/client/components/vision2/camera/world_line.ts
TrentHouliston: Should all of these functions get docstrings? I have noticed that most of NUsight2 doesn't have them

src/client/components/vision2/camera/line_projection.ts
```diff
@@ -0,0 +1,105 @@
...
+        : new THREE.Vector3(-axis.y, axis.x, 0).normalize(),
+    )
+    return this.coneSegment({ axis, start, end: start, color, lineWidth })
+  }
+
+  /** Draws a segment of a plane projected to infinity in world space. */
+  planeSegment(segment: { axis?: Vector3, start: Vector3, end: Vector3, color?: Vector4, lineWidth?: number }) {
+    return this.coneSegment({
+      ...segment,
+      axis: Vector3.fromThree(
```
TrentHouliston: Still need to use the `segment.axis` here if provided. The problem is that when you have a 360 or 180-degree angle, crossing them gives no normal vector to project the plane on. So either we need to find a way around that or provide an axis here.

src/client/components/vision2/camera/model.ts
```diff
@@ -0,0 +1,54 @@
...
+  @observable.ref projection: Projection
+  @observable.ref focalLength: number
+  @observable.ref centre?: Vector2
+  @observable.ref distortionCoeffecients?: Vector2
+
+  constructor({ projection, focalLength, centre, distortionCoeffecients }: {
+    projection: Projection,
+    focalLength: number,
+    centre?: Vector2,
+    distortionCoeffecients?: Vector2
```
TrentHouliston: Should these be optional? Are they defaulted somewhere?

src/client/components/vision2/camera/line_projection.ts
```diff
@@ -0,0 +1,105 @@
...
+import { rawShader } from '../../three/builders'
+import { shaderMaterial } from '../../three/builders'
+import { Canvas } from '../../three/three'
+
+import { Lens } from './model'
+import fragmentShader from './shaders/line_projection.frag'
+import vertexShader from './shaders/line_projection.vert'
+
+export interface ConeSegment {
+  /** The camera space central axis of the cone. Only needed for a plane if the start and end are (anti-)parallel. */
```
TrentHouliston: This comment now makes no sense
```suggestion
  /** The camera space central axis of the cone. */
```

src/client/components/vision2/camera/model.ts
BrendanAnnable: Yes! They are defaulted in the material: https://github.com/NUbots/NUsight2/pull/312/files#diff-59b40caddcdba31574cf9e3beea3bbc0R80

Otherwise I have to default them in every story, which seems unnecessary since they seem optional "corrections".

PR #315, opened 2020-02-21
Integrate vision v2

BrendanAnnable:
https://nusight-annable-vision-gdht8gr.herokuapp.com/vision2


src/client/components/vision2/camera/model.ts
```diff
@@ -79,6 +79,21 @@ export class CameraModel {
...
+    this.name = that.name
+    this.image = that.image
+    this.params.copy(that.params)
+    if (this.greenhorizon && that.greenhorizon) {
+      this.greenhorizon.copy(that.greenhorizon)
+    } else {
+      this.greenhorizon = that.greenhorizon
+    }
+    this.balls = that.balls // TODO: deep copy?
+    this.goals = that.goals // TODO: deep copy?
```
TrentHouliston: I don't know... should you deep copy?

src/client/components/vision2/install.tsx
```diff
@@ -0,0 +1,34 @@
...
+  Menu,
+}: {
+  nav: NavigationConfiguration
+  appModel: AppModel
+  nusightNetwork: NUsightNetwork
+  Menu: ComponentType
+}) {
+  const model = VisionModel.of(appModel)
+  nav.addRoute({
+    path: '/vision2',
```
TrentHouliston: When we merge this PR, do we just want to replace vision1 with it?

src/client/components/vision2/camera/model.ts
BrendanAnnable: Maybe! There are potential performance wins for reusing meshes. Currently this blows them all away.

Not critical either way as the expensive shaders are still cached.

src/client/components/vision2/styles.css
```diff
@@ -0,0 +1,14 @@
+.vision {
+  display: flex;
+  flex-direction: column;
+  flex-grow: 1;
+  flex-shrink: 1;
+  text-align: center;
+}
+
+.cameras {
```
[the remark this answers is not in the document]
BrendanAnnable: > The canvas should maintain the aspect ratio of the image if possible

It does! The black bars are just from a css background color.

> Also, it would be good to be able to see more than a single image at a time.

Indeed. I think a some reasonable camera layout algorithm is a prerequisite to merging this PR.

src/client/components/vision2/install.tsx
BrendanAnnable: Yep! Once this is ready I'll raise a PR on top of this doing the rename/replacement, then merge them both together into master.

PR #317, opened 2020-02-23
Delete and recreate yarn.lock

TrentHouliston:
Storybook would not compile for me until I did this


conf/storybook/webpack.config.ts
```diff
@@ -23,7 +23,6 @@ export default ({ config: storybookConfig }: { config: webpack.Configuration })
       ...(config.plugins || []).filter(p => !(
           p instanceof HtmlWebpackPlugin // Storybook handles page generation.
           || p instanceof CopyWebpackPlugin // Avoids overwriting index.html.
-          || p instanceof ProgressBarPlugin // Storybook already has a progress plugin.
```
BrendanAnnable: This was meant to prevent the double progress bar when starting storybook, but it doesn't seem to be working anyway, so...

BrendanAnnable: oh, you should also delete the import

PR #327, opened 2020-07-06
Add smarter networking packet rate control

BrendanAnnable:


src/server/nuclearnet/web_socket_proxy_nuclearnet_server.ts
```diff
@@ -200,22 +203,52 @@ class PacketProcessor {
...
           if (!isDone) {
-            this.outgoingPackets--
+            // Update our performance tracking information
+            stats.active -= 1
+            if (processingTime != null) {
+              // Calculate network latency
+              const latency = this.clock.performanceNow() - start
+              const n = stats.active
+              stats.latency = (stats.latency * n + latency) / (n + 1)
+              stats.processingTime = (stats.processingTime * n + processingTime) / (n + 1)
```
TrentHouliston: This is supposed to be an exponential average where n is some constant which defines the inertia of the moving average. It was previously 10 in the prior PR as a magic number

src/server/nuclearnet/web_socket_proxy_nuclearnet_server.ts
```diff
@@ -200,22 +203,52 @@ class PacketProcessor {
...
+            }
             isDone = true
             this.maybeSendNextPacket()
           }
         }
-        this.outgoingPackets++
-        this.socket.volatileSend(event, packet, done)
+        // this.socket.volatileSend(event, packet, done)
```
TrentHouliston: Does `volatileSend` do anything anymore?

src/server/nuclearnet/web_socket_proxy_nuclearnet_server.ts
```diff
@@ -200,22 +203,52 @@ class PacketProcessor {
...
+        this.socket.send(event, packet, done)
+        stats.lastSent = this.clock.performanceNow()
         this.clock.setTimeout(done, this.timeout)
       }
     }
   }
+
+  private canSendEvent(event: string): boolean {
+    const stats = this.getStatsForEvent(event)
+    const maxRate = stats.processingTime / stats.latency
```
TrentHouliston: I think if we are sending based on a rate, we can just use processing time alone. We want to send packets at a rate that matches the processing time on the other end

src/server/nuclearnet/web_socket_proxy_nuclearnet_server.ts
BrendanAnnable: Is there something here I can link to? https://en.wikipedia.org/wiki/Moving_average

TrentHouliston: https://en.wikipedia.org/wiki/Exponential_smoothing would do it

PR #331, opened 2020-10-07
Fix crash in <GridLayout> when it has no children

JosephusPaye:
`bestFitCols()` assumes there's at least one child, and crashes when there's none:

### DIFF
--- a/src/client/components/vision/grid_layout/grid_layout.tsx
+++ b/src/client/components/vision/grid_layout/grid_layout.tsx
@@ -44,7 +44,7 @@ export class GridLayout extends Component<{
   @computed
   get gridTemplate() {
     const n = React.Children.count(this.props.children)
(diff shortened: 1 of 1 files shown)


src/client/components/vision/grid_layout/grid_layout.tsx
```diff
@@ -44,7 +44,7 @@ export class GridLayout extends Component<{
   @computed
   get gridTemplate() {
     const n = React.Children.count(this.props.children)
-    if (!this.width || !this.height) {
+    if (!this.width || !this.height || n < 2) {
```
JosephusPaye, line 47: Returns `1fr`(single row, single column) when count of children is zero or one. There's no need to call `bestFitCols()` in this case.

TrentHouliston (approved):
LGTM!

PR #333, opened 2020-10-28
Fix projection for rectilinear cameras so they don't get mirrors

TrentHouliston:
Currently, if the ray is behind the camera for rectilinear cameras, we get a negative R value out from tan.
Technically if the ray is behind the camera we shouldn't see it at all.
This fixes the issue by clamping the angle so we can't get negative r values


src/client/components/vision/camera/shaders/line_projection.frag
```diff
@@ -144,7 +146,7 @@ float rectilinearTheta(float r, float f) {
  * @return the distance from the optical centre when the point is projected onto the screen
  */
 float rectilinearR(float theta, float f) {
-  return f * tan(theta);
+  return f * tan(clamp(theta, 0.0, M_PI_2));
```
JosephusPaye, line 149: A comment like what you have in the PR description would be good to know why it's clamped

JosephusPaye, line 149:
```suggestion
  // Clamp the angle so we can't get negative R values, which happens if the ray is behind the camera for rectilinear cameras
  return f * tan(clamp(theta, 0.0, M_PI_2));
```

JosephusPaye (changes requested):
Comment

PR #334, opened 2020-10-28
Fix `unproject()` in line projection shader

JosephusPaye:
As [reported by Aaron](https://nubotsteam.slack.com/archives/C3T1J8SGL/p1603755858002000) in Slack


TrentHouliston (approved):
Sure!

BrendanAnnable (commented):
Is this wrong as well then?

https://github.com/NUbots/NUsight2/blob/38af0e3ed16caa8ba13efa2f83144b39d8057808/src/client/components/vision/camera/stories/camera.stories.tsx#L135-L136

TrentHouliston:
> Is this wrong as well then?
> 
> https://github.com/NUbots/NUsight2/blob/38af0e3ed16caa8ba13efa2f83144b39d8057808/src/client/components/vision/camera/stories/camera.stories.tsx#L135-L136

Yep! 
Although I think in general that story doesn't deal with any of the calibration parameters properly.
Since it doesn't deal with distortion technically there is no distorted r to use

PR #335, opened 2020-11-18
Allow running in Docker on Windows and macOS

JosephusPaye:
Pairs with https://github.com/NUbots/NUbots/pull/341.

This adds a simple docker-compose file for running NUsight in a docker container to get it to talk to NUbots in a container on Windows.

Using `docker-compose` (instead of just `docker`) allows for creating the container on demand without a Dockerfile + `docker build`.

It also adds a shell script to make it easier to run the docker-compose commands. I considered adding the commands as scripts to `package.json`, but `yarn` reports two spurious errors on SIGINT exit.

---

To be used as follows (proper NUbook documentation coming soon), all commands run in WSL on Windows:

### Setup

- Create the network: `docker network create nuclearnet`
- Make the docker script executable: `chmod +x ./docker`

### Build

- Install NUsight deps in the container `./docker yarn`
- Build NUsight for prod in the container `./docker yarn build`

### Run

- Run something in NUbots and connect it to the network: 
  ```sh
  ./b run keyboardwalkfake --network nuclearnet
  ```

- Inspect the network to find the NUbots container's IP address, X:
  ```sh
  docker inspect --format='{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' nubots_generic
  ```

- Run NUsight with the NUbots container address:
  ```sh
  ./docker yarn prod --address X
  ```

### DIFF
--- a/docker
+++ b/docker
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+
+# Run `yarn ...` on `./docker yarn ...`
+if [ "$1" == "yarn" ]; then
+    shift 1
+    docker-compose run \
+        -p 3000:3000 \
+        -p 9090:9090 \
+        nusight \
+        yarn "$@"
+
+# Open a bash shell in the container on `./docker shell`
+elif [ "$1" == "shell" ]; then
+    shift 1
+    docker-compose run \
+        -p 3000:3000 \
+        -p 9090:9090 \
+        nusight \
+        bash
+
+# Otherwise show usage information
+else
+    echo "usage:"
+    echo "./docker yarn [command]   # run the given yarn command in a docker container"
+    echo "./docker shell            # open a bash shell in a docker container"
+fi
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,8 +4,12 @@ services:
     image: node:lts
     volumes:
       - .:/var/www/NUsight2
+      - node_modules_volume:/var/www/NUsight2/node_modules
     working_dir: /var/www/NUsight2
 networks:
   default:
     external:
       name: nuclearnet
+volumes:
+  node_modules_volume:
+    driver: local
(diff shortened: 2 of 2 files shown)


ysims (approved):
I've tested this and it all works. Script looks fine.

BrendanAnnable (commented):
LGTM if LGTTrent, can we add documentation somewhere?

TrentHouliston (changes requested):
A few comments based on when I did this before.
I would recommend mounting a volume over the node_modules folder so you can have both the local and docker versions still work.
Commit a .gitignore file in the node_modules folder to keep it anchored.

Also I would recommend rather than having to run a script to find the address of the nubots system, just using the multicasting or broadcasting that is built into NUClear as it works just fine over the docker network

JosephusPaye:
It currently mounts the entire NUsight2 folder, including `node_modules` in the container.

How do we use the multicasting/broadcasting?

TrentHouliston:
> It currently mounts the entire NUsight2 folder, including `node_modules` in the container.

You need to overmount a new volume over the `node_modules` folder that way the host has a different copy of the files

> How do we use the multicasting/broadcasting?

You'll have to test it but I'm pretty sure that just using the default multicast address that NUsight2 defaults to should work.
Within the docker network, it should sort itself out without you having to set anything special.
Really it should work on the normal robots too.

JosephusPaye:
Is there something we need to do on the NUsight side for multicasting?

I just tried and it doesn't work without an explicit `--address`. Probably because it tries to connect to the default `10.1.255.255` that we use in the lab?

JosephusPaye:
Documented in NUbook: https://nubook.netlify.app/guides/main/getting-started#get-build-and-run-nusight

TrentHouliston (approved):
@JosephusPaye promised that he would make more PRs to fix the multicast addressing!

PR #336, opened 2021-01-12
Fix 3rd-party CSS files in development

JosephusPaye:
The `include` and `exclude` loader checks (see below) for CSS files in `node_modules/` currently don't work in development, as `__dirname` is `nusight/dist/` when the webpack config file is built, not `nusight/`.

https://github.com/NUbots/NUsight2/blob/ed96ed4d23aa909e93a5ae8e72f037e7826e01d0/webpack.config.ts#L59-L62

https://github.com/NUbots/NUsight2/blob/ed96ed4d23aa909e93a5ae8e72f037e7826e01d0/webpack.config.ts#L92-L98

### DIFF
--- a/conf/storybook/webpack.config.ts
+++ b/conf/storybook/webpack.config.ts
@@ -9,6 +9,7 @@ export default ({ config: storybookConfig }: { config: webpack.Configuration }) 
     mode: 'development',
     context: path.resolve(path.join(__dirname, '..', '..', 'src')),
     sourceMap: 'eval-source-map',
+    rootDir: path.join(__dirname, '..', '..'),
   })
   return {
     ...config,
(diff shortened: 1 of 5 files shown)


webpack.config.ts
```diff
@@ -59,7 +59,11 @@ export function getClientConfig({
         // local css
         {
           test: /\.css$/,
-          exclude: [path.resolve(__dirname, 'node_modules')],
+          exclude: [
+            __dirname.endsWith(path.sep + 'dist') // __dirname is `[..]/dist` when built server is run in dev
```
BrendanAnnable, line 63: I don't understand. My understanding is that `__dirname` is the directory the current file is in. This file is top-level and doesn't move?

JosephusPaye: After updating `webpack.config.ts` like so,

Screenshot of updated webpack config file

Running `yarn dev`:

```
yarn run v1.22.5
$ yarn start:virtual
$ ts-node src/server/dev-server.ts --virtual-robots
[...]
Starting server
{ __dirname: '/home/jpaye/horus/nusight2/dist' }
```

Running `yarn build`:

```
yarn run v1.22.5
$ ts-node build_scripts/build.ts --production
{ __dirname: '/home/jpaye/horus/nusight2' }
```

Running `yarn storybook`:

```
yarn run v1.22.5
$ node conf/storybook/dev.js
[...]
{ __dirname: '/home/jpaye/horus/nusight2' }
```

---

I think the issue is that `webpack.config.ts` gets built into `dist/` when the server is built during dev, then run from there when the client is built.

webpack.config.ts
```diff
@@ -95,7 +99,11 @@ export function getClientConfig({
         */
         {
           test: /\.css$/,
-          include: [path.resolve(__dirname, 'node_modules')],
+          include: [
+            __dirname.endsWith(path.sep + 'dist') // __dirname is `[..]/dist` when built server is run in dev
+              ? path.resolve(__dirname, '..', 'node_modules')
```
BrendanAnnable, line 104: It probably doesn't make sense for this file to understand this relationship.

How about passing in directories as arguments to these functions (`getServerConfig`/`getClientConfig`)?

We can remove all references to `__dirname` that way.

BrendanAnnable (commented):
Ah I understand now. Could we inject instead of more hardcoding?

webpack.config.ts
JosephusPaye: Yeah, that's a better approach. I'll update this PR tomorrow.

conf/storybook/webpack.config.ts
```diff
@@ -18,19 +19,23 @@ export default ({ config: storybookConfig }: { config: webpack.Configuration })
       rules: config.module && config.module.rules,
     },
     plugins: [
-      ...storybookConfig.plugins || [],
```
JosephusPaye, line 21: Prettier formatted this on save.

webpack.config.ts
```diff
@@ -11,13 +11,15 @@ export type ConfigOptions = {
...
   transpileOnly?: boolean
+  rootDir: string
 }
 
 export function getClientConfig({
   mode,
   context,
   sourceMap,
   transpileOnly,
+  rootDir,
```
BrendanAnnable, line 22: Perhaps default its value (and make it optional), such that we don't have to pass it in most of the time?

```suggestion
  rootDir = __dirname,
```

BrendanAnnable (dismissed):
LGTM!

